PR #31, opened 2025-05-15
Corpan

skyl:


corpan/corpan-app/public/tauri.svg
skyl, line 1: rm

corpan/corpan-app/src-tauri/icons/icon.icns
skyl, line 1: Not sure that we shouldn't only put a couple of reference icons into version control since `tauri icon` seems to do the business with just 1 icon? So, maybe all these others should be gitignore'd ... if we're going full full monorepo, we should clean up the bloat and history of the binaries/icons/etc ...

PR #61, opened 2025-06-13
Add text size adjustment

0bkevin:
Allow users to select the text size and adjust the text size across all the app. 

The text size adjust across all the app text, to match the selected option of the user. 

image
image


0bkevin:
@skyl, check this out.  Still i have to make the iOS text size adjustment is supported on the app, but this is an in-app text size adjust.

PR #66, opened 2025-06-26
Start fixing UI/UX 

0bkevin:
- [x] The X in the settings should float when scrolling so you can leave without scrolling back to the top. X should be fixed.
- I just add a scrolled area wrapping all the settings, letting the title, subtitle and close button fixed.


skyl:
conflicted adn we are probably going to rtock the boat pretty hard and do this some other way .

PR #78, opened 2025-07-25
Add Vietnamese and Farsi language entries to database

Umanistan:
Added Farsi and Vietnamese entries via Django shell for backend support.


skyl:
#79 got vietnamese

PR #84, opened 2025-08-05
feat(Corpan): add Persian

skyl:


skyl:
Persian plays twice on iOS and not at all on Android but I got excited and released to production so now we must go on

PR #85, opened 2025-08-06
Improve i18n in corpan with react-i18next

perryd01:
- eject in app translations from the store
- create separate translation files validated by a json schema


perryd01:
i think its more or less done

PR #95, opened 2025-09-18
feat: autplay

0bkevin:
Support auto play functionality in the app.


skyl:
we should probably go ahead and merge this: https://github.com/corpora-inc/encorpora/pull/94/files - there is a big refactor of the state there that new PRs will need to merge in. That branch still has some left to do, maybe I can finish tonight.

skyl:
Thanks for this, Kevin — the autoplay UX (play/stop toggle, per-sentence delay slider in Settings, tap-guarding the speak buttons while autoplaying) is a genuinely nice feature and the implementation was clean against the code as it stood.

Unfortunately `main` has moved a long way since this branched (~10 days), and every file this PR touches has been structurally rewritten, so it can no longer be merged as a conflict resolution — it would need a full re-implementation:

- **`MainExperience.tsx`**: the render loop no longer uses `textByLang[code]` / `createVoiceTTS(...)`. It now renders a `TranslationBlock` component fed by `pickText(lookup.textByDbCode, uiCode)` and a shared `speak(uiCode, txt)` helper. The two core symbols the autoplay loop is built on — `textByLang` and `createVoiceTTS` — no longer exist in this file on `main`. The nav also switched `fetchRandomEntry` → `handleRandom` and added `ids.length`-based counters.
- **`store/settings.ts`**: migrated to a stacks-based model (`stacks` / `activeStackId` / `deriveFrom`); the flat reset/rate block the diff edits is gone.
- **`util/speak.ts`**: `speak()` gained a `voiceId` param and per-language fallback shims.
- **`pnpm-lock.yaml`**: the app switched from pnpm to npm (`package-lock.json`), so the lockfile changes here are obsolete, and the `react-i18next` / `i18next` deps you added are already on `main`.

Closing as superseded by the refactor rather than leave it to rot. The feature is still wanted — if you'd like to re-land it on current `main`, the autoplay state machine should drive the new `speak(uiCode, txt)` helper and iterate `displayedLanguages` via `pickText(lookup.textByDbCode, ...)`, the disable-during-autoplay wiring moves into `TranslationBlock`, and `autoplayDelayMs` slots cleanly into the new settings store. Happy to review a fresh PR. The locale keys (`settings.autoPlayDelay`, `settings.betweenSentences`) are reusable as-is.

PR #100, opened 2025-10-08
Feat/scroll: Allow users to navigate through sentences by scrolling vertically or horizontally

0bkevin:
Allow users to navigate through sentences by scrolling vertically or horizontally. Users can now swipe on touch devices or use mouse/trackpad scrolling to move between sentences, providing a more intuitive and modern navigation experience


skyl:
For some of these it might just be the squash merge that could be fixed with eg `git merge main -X theirs `.

We need to get things in main faster so they don't stay out and cause conflicts.

skyl:
merged in #123

PR #269, opened 2026-06-06
0.17.0 — Corpan City + on-device ASR dictation + multiplayer

skyl:
# 0.17.0 — Corpan City, on-device ASR, multiplayer

The consolidated 0.17.0 release. App bumped **0.16.2 → 0.17.0**. `main` is a strict ancestor — clean fast-forward, all CI green.

## 🏙️ Corpan City (the World Plaza pack, renamed)
A living HD-2D language city you explore, with AI characters and real players.
- **Renamed** World Plaza → **Corpan City** (id `corpan_city`, pack v0.1.0); catalog entry gated **minAppVersion ≥ 0.17.0**, localized name/blurb (51 locales) + an in-engine 512² store icon.
- **Believable NPCs:** venue-typed personas (clinic→doctor, café→barista…), modern Corpan City roles, replies in the **target language**, voice = current target (stickiness clears on entry/stack change).
- **Localized quests** (titles/descriptions/names, 46 langs) that flip native↔target with immersion.
- **Google-Maps-style map** (declutter → tappable pins, localized popovers, route/course, no-quest), phone-OS shell, **re-onboarding** (Profile app), badges/logo polish.
- **Robustness:** safe-spawn everywhere (no more fountain trap), resume-at-exit per pair, roof-fade, 60 FPS, radio stops on exit.

## 🎤 On-device ASR dictation (native-first)
- `tauri-plugin-asr-native` (Apple SpeechAnalyzer/SFSpeechRecognizer + Android SpeechRecognizer) + `corpan-asr-contract` + `host.asr`/`host.models` seam.
- Optional **speech-to-text mic** in the Corpan City NPC chat **and** Tutomaton — zero download, on-device, private, degrade-safe (keyboard floor where unsupported). Device-validated on iPad.

## 🌐 Multiplayer
- Colyseus presence server deployed to **AWS App Runner** (scale-to-near-zero), live URL baked into the encorpora.io catalog — real players together in Corpan City.

## ✨ Onboarding
- Rewritten welcome screen — "you're an early adopter": alpha / cutting-edge / works-offline / no-data-collection / eco framing (no poor-mouth), localized into all ~50 languages.

## 🧱 Foundation & misc
- Babylon **9.11** upgrade; Tutomaton prompt simplification (~50 langs) + streaming TTS / voice-preferences; assorted host + plugin fixes.

→ TestFlight after merge.


skyl:
## 🚀 Corpan City multiplayer is live

**Presence server:** `wss://fatmqwnfjv.us-east-2.awsapprunner.com` (AWS App Runner, us-east-2, always-warm).

### What landed on this branch (since last review)

- **`.github/workflows/hover-runner-pages.yml`** — added world-plaza build → `scripts/pack.mjs` → `world-plaza.zip` → io/out copy steps, with `VITE_WP_SERVER_URL` baked in at build time. Once this merges to `main`, the GH Pages deploy publishes `encorpora.io/corpan/packs/world-plaza.zip` and the existing `corpan_city` entry in `web/data/packs.json` (gated `minAppVersion: 0.17.0`) appears in `catalog.json` — so any app 0.17.0+ installs Corpan City with multiplayer ON.
- **`corpan/packs/world-plaza/scripts/pack.mjs`** — new (mirrors `world-radio/scripts/pack.mjs`); zips `manifest.json + dist/`. Content JSON is inlined into `dist/app.js` by Vite, so the zip is code-only.
- **`corpan/packs/world-plaza/server/Dockerfile`** — two crashloop fixes from the first deploys:
  1. `npm ci --include=dev` — without it, `NODE_ENV=production` had npm omit `tsx` (the runtime), and the container died with `Cannot find module '/app/node_modules/tsx/dist/cli.mjs'`.
  2. `--tsconfig /app/server/tsconfig.json` + `TSX_TSCONFIG_PATH` — without it, tsx's default tsconfig discovery (CWD `/app`) misses the `@world-plaza/contracts` path alias and dies with `Cannot find package '@world-plaza/contracts'`.
- **`corpan/packs/world-plaza/server/package-lock.json`** — regenerated; was missing `@pm2/io` + transitives, blocking `npm ci` with EUSAGE.
- **`corpan/packs/world-plaza/server/DEPLOY.md`** — §0 corrected: `corpan-publisher` is S3/CloudFront-only and cannot do this deploy. Use `AWS_ACCESS_KEY_ID/SECRET` (the `terraform-admin` user from `encorpora/.env`).
- **`corpan/packs/world-plaza/CHANGELOG.md`** — `[Unreleased]` entry for the catalog ship.

### Smoke test (server-side M1 wow proven)

```
$ curl -s https://fatmqwnfjv.us-east-2.awsapprunner.com/matchmake/joinOrCreate/plaza -H 'content-type: application/json' -d '{}'
{"room":{"clients":1, ... "roomId":"2nZ1g9iVI"}, "sessionId":"Lxi8CTSzU"}

$ curl -s ... (second client)
{"room":{"clients":2, ... "roomId":"2nZ1g9iVI"}, "sessionId":"8mqkWfBFL"}
```

Two requests → same `roomId` → `clients: 2`. That's the matchmaking spine working end-to-end against the live server.

### Path to TestFlight

1. **Squash-merge this PR.** GH Pages deploys, `world-plaza.zip` lands at `encorpora.io/corpan/packs/world-plaza.zip`, `catalog.json` gains the `corpan_city` entry (gated ≥ 0.17.0), and the baked URL is the live wss.
2. **macOS build:** `cd corpan-app && npm run tauri ios build`, archive in Xcode, upload to TestFlight (the iOS Tauri project gen step needs Xcode; can't be done from this Linux box).
3. **Install on two devices** → enter Corpan City → walk both avatars near each other → the nearby-pip lights, the safe profile card opens, real-time movement syncs. M1 wow.

### Cost
~$45–60/mo for the always-warm 0.25 vCPU / 0.5 GB instance (`min_size = 1` because presence needs a live socket). Scales to 3 under load.

### Rollback / fresh deploys
`server/DEPLOY.md` is up to date. New tag → docker push → `terraform apply -var plaza_server_image_tag=<tag>` → update `WP_SERVER_URL` repo variable (or this workflow line) if AWS reassigns the URL.

@skyl — ready for your friend. 🎮

skyl:
## Release review: request changes before production merge

I reviewed the release-critical paths across the total diff, generated the catalogs, ran the app's real catalog filter, built/packaged Corpan City, exercised the live App Runner matchmaking endpoint, and ran the available tests/typechecks.

### Blocking findings

1. **Every Corpan City client joins multiplayer with the same `playerId`, so routed player interactions break once multiple users connect.**

   [`game.ts` passes `playerId: "player-local"` to multiplayer](https://github.com/corpora-inc/encorpora/blob/ae6f3695564c304d8305d558c5ae418b4eabb7c0/corpan/packs/world-plaza/src/game.ts#L1090-L1093), and onboarding also seeds identities with that same value. The server stores one session per player ID in [`byPlayerId`](https://github.com/corpora-inc/encorpora/blob/ae6f3695564c304d8305d558c5ae418b4eabb7c0/corpan/packs/world-plaza/server/src/PlazaRoom.ts#L125-L145), then uses that map for profiles/invites/chat/trade. Each new join overwrites the previous `player-local` mapping, so those interactions route to the most recently joined client. Presence movement still appears to work because it uses Colyseus session IDs, which is why the current matchmaking smoke test does not catch this.

   There is already a `loadOrMintPlayerId()` helper. We should wire that through onboarding and `initMultiplayer`, then add a two-client routing test.

2. **Free-text multiplayer chat is not actually mediated before crossing devices and can be displayed unmoderated.**

   [`composeOutbound` puts the user's raw text on the wire](https://github.com/corpora-inc/encorpora/blob/ae6f3695564c304d8305d558c5ae418b4eabb7c0/corpan/packs/world-plaza/src/multiplayer/mediatedChat.ts#L54-L74), the server [forwards it as-is](https://github.com/corpora-inc/encorpora/blob/ae6f3695564c304d8305d558c5ae418b4eabb7c0/corpan/packs/world-plaza/server/src/PlazaRoom.ts#L326-L344), and the failure fallback [shows the original text while marking it allowed/OK](https://github.com/corpora-inc/encorpora/blob/ae6f3695564c304d8305d558c5ae418b4eabb7c0/corpan/packs/world-plaza/src/multiplayer/mediatedChat.ts#L76-L96). If the recipient has no LLM, the model is busy, inference fails, or JSON parsing fails, links/contact info/abuse are shown unchanged. That contradicts the repeated “never raw UGC” guarantee and is especially risky because the public catalog marks Corpan City kid-friendly.

   For the first production launch, disable free-text peer chat or make the fallback block/curate the message. Presence, profiles, and curated phrase-card interactions can ship independently.

### Will the latest Corpan City go live to 0.17 users?

**Yes, after merge to `main` and a successful `Deploy Encorpora to GitHub Pages` run.** The workflow builds the merged Corpan City source with the live server URL baked in, packages `world-plaza.zip`, and publishes it with `catalog-v3.json`.

Verified with the app's actual catalog parser/filter:

- Corpán `0.16.99`: does **not** receive Corpan City
- Corpán `0.17.0`: receives `corpan_city@0.1.0`
- Corpán `0.17.1`: receives `corpan_city@0.1.0`
- Catalog localization: name/description/tagline complete for all 51 app locales
- The live App Runner matchmaking endpoint returned `200` and created a `plaza` room
- Apps may retain the previous catalog for up to one hour before refreshing

Correction to the earlier PR comment: Corpan City is intentionally excluded from the older `catalog.json` via `v1Listed: false`; it ships through `catalog-v3.json`, which is what enforces the `minAppVersion: 0.17.0` gate.

One update caveat: the pack version remains `0.1.0`. This is fine for the first public publish, but anyone who already installed a pre-release `corpan_city@0.1.0` will not be offered the rebuilt pack because pack updates require the remote version to increase. Bump the pack version if those testers must receive it automatically.

### Verification performed

- PR checks green: Corpán app typecheck/build, web/io build, Terraform fmt/validate
- Corpan City: typecheck; **668 tests passed**; translation check passed; production build and ZIP integrity passed; live WSS URL confirmed in the built IIFE
- Corpan City server: typecheck passed; no server test script exists
- Corpán native host: `cargo check --locked` passed
- Catalog generation and compatibility routing passed

Process risk: this PR is 1,037 files / ~240k additions and GitHub cannot render the complete PR diff. Corpan City itself also has no PR CI job; its build currently runs only in the post-merge Pages deployment. The targeted local verification above is green, but adding a Corpan City test/typecheck/build PR gate would materially reduce release risk.

skyl:
## Follow-up: multiplayer blockers fixed and verified locally

The two release blockers from the review are now implemented in the `world-plaza` working tree. These changes still need to be committed and pushed before they are part of this PR.

- **Stable routing identity:** Corpan City now calls the existing `loadOrMintPlayerId()` at startup, binds onboarding/default/saved identity to that per-install anonymous ID, and passes it to `initMultiplayer`. The server also refuses duplicate/malformed requested IDs instead of allowing a later connection to overwrite another player route.
- **Two-sided mediated chat:** raw authored text stays local. The sender local LLM preserves intent while cleaning/moderating before transmission; the recipient local LLM independently cleans again and translates into the recipient learning language, with native-language meaning. Missing/busy/failed/invalid model paths use fixed harmless fallback messages and never display the received intent. Deterministic contact/link checks run after both model passes.
- **Trusted server route:** chat delivery stamps the live sender, resolved recipient, and recipient target language instead of trusting caller-supplied routing fields.
- **Tests:** added stable-ID coverage, two-sided mediation/failure/contact-info coverage, and a live three-connection interaction QA proving a duplicate player ID cannot steal routing and a forged chat sender is replaced.

Verification completed:

- Corpan City full suite: **67 files / 669 tests passed**
- client typecheck + server typecheck passed
- all 46 pack locales complete
- live local multiplayer interaction QA passed: profile, duplicate-ID defense, invite, trusted chat route, peer result, trade
- production build, pack, and ZIP integrity passed
- `git diff --check` passed

Release behavior remains: after this fix is committed/pushed, merge to `main`, and a successful Pages deploy, the rebuilt Corpan City client will be offered through `catalog-v3.json` to Corpán **0.17.0+** users. The client-side identity and two-sided mediation fixes remain wire-compatible with the currently deployed App Runner server. The new server-side duplicate-ID/trusted-route hardening requires a separate App Runner image redeploy to become live.

skyl:
Review pass on merge/TestFlight readiness:

I would hold merge for a short follow-up. CI is green and the shared presence backend is live/healthy, but I see a few things that can bite us immediately when trying 0.17.1 through TestFlight.

Findings:

1. Teletron is in the pack catalog, but I don't see the Pages workflow publishing its zip artifact. `web/data/packs.json` points Teletron at `/corpan/packs/teletron.zip` (`web/data/packs.json:163-164`), while `.github/workflows/hover-runner-pages.yml` packages/copies Corpan City and Tutomaton (`.github/workflows/hover-runner-pages.yml:181-266`) but has no Teletron build/package/copy step. The live URL 404s today, which is expected pre-merge, but without the workflow step it likely stays a broken catalog entry after deploy.

2. Plus-unlimited for Teletron is not wired through the app host yet. Teletron checks `initial.isPlus`, `initial.entitlement.plus`, or `globalThis.__CORPAN_PLUS` (`corpan/packs/teletron/src/main.ts:140-145`), but `ContentPackHost` currently mounts packs with only `stackConfig` and optional entry info (`corpan/corpan-app/src/contentPacks/ContentPackHost.tsx:440-444`). Result: Plus users will likely hit the same daily limit as free users. If we want to ship Teletron to everyone with "Plus gets unlimited", this needs a host-side entitlement bridge or the catalog/UX should call it a known limitation for the first TestFlight.

3. There is a compiled Mach-O helper binary tracked in the PR: `corpan/scripts/dev/ipad/ipad-record`. The Swift source is also present (`corpan/scripts/dev/ipad/ipad-record.swift`), so I would remove the binary from git and build it locally/on demand. This is not a product blocker, but it is a clean-PR blocker in my view.

Impact on <0.17 users:

This part looks mostly safe. The new packs are gated with `minAppVersion: 0.17.0` and `v1Listed: false`; legacy Tutomaton remains available to 0.16 via the 0.3.2 entry with `minAppVersion: 0.16.0` / `maxAppVersion: 0.16.99` (`web/data/packs.json:43-63`). The app catalog code filters on min/max app version, and the v1 web catalog should exclude `v1Listed: false` packs. I would still verify the generated `catalog.json` and `catalog-v3.json` from the Pages build before shipping, but the intended compatibility shape is right.

Other checks:

- PR checks are passing: app build, web/io build, terraform fmt/validate.
- Secret scan over changed files did not find literal AWS keys/private keys/tokens; hits were placeholders/env var names/code variable names.
- Live shared presence endpoint is healthy, and Colyseus room joins/profile round-trip were tested for plaza/teletron.
- Terraform plan for the deployed EC2 presence setup is clean.

Recommendation:

Not quite ready to merge if the goal is "build 0.17.1 on another machine, put it on TestFlight, and try the new packs from the CDN catalog." I would fix the Teletron artifact publishing step, wire or explicitly defer the Plus entitlement bridge, and remove the tracked compiled binary. After that, this looks like a reasonable 0.17.1 TestFlight candidate.

skyl:
Follow-up: I fixed the Pages artifact gap and cleaned up the workflow naming.

Changes pushed:

- `5ca53979` adds Teletron to the GitHub Pages pack pipeline: install, build, zip, and copy to `web/io/out/corpan/packs/teletron.zip`. It also updates the Corpan City/Teletron build-time fallback to the current shared presence endpoint (`wss://presence.3-142-26-37.sslip.io`) while preserving the `vars.WP_SERVER_URL` override.
- `987ac6a1` renames `.github/workflows/hover-runner-pages.yml` to `.github/workflows/deploy-pages.yml` so future work doesn’t have to decode the old Hover Runner-specific filename.
- `987ac6a1` also adds a lightweight changed-pack CI job. When `corpan/packs/**` changes, it discovers touched pack folders with `package.json` and runs install, optional typecheck, and optional build for those packs. No new deploy secrets or AWS behavior.

Updated readiness read:

- The Teletron CDN artifact concern from my review should now be addressed by the Pages workflow.
- <0.17 user compatibility still looks safe via the catalog min/max version and `v1Listed: false` gating.
- Remaining product caveat is still Plus entitlement bridging into content packs: Teletron can enforce the free daily limit client-side today, but “Plus gets unlimited” needs the host to pass entitlement state into pack initial state/API before we can truly validate that behavior in TestFlight.
- I still recommend removing the tracked compiled `corpan/scripts/dev/ipad/ipad-record` binary before merge unless we intentionally want that artifact in git.

Local verification after the workflow changes:

- YAML parsed for both workflows.
- Teletron production build passes under Node 22 locally; the raw `npm run build` on this machine still fails because the shell Node is 18.19.1, while GitHub Actions installs Node 20.
- Teletron zip command produces `manifest.json`, `dist/app.css`, and `dist/app.js`, and those generated files remain gitignored.

skyl:
Final follow-up after the last round of fixes:

I think this PR is now ready to merge for a 0.17.1 TestFlight candidate.

What changed since the earlier review:

- Teletron is now deployed by the GitHub Pages pack pipeline: built, zipped, copied to `/corpan/packs/teletron.zip`, and its manifest/dist/avatar are copied under `/corpan/packs/teletron/`.
- The old `hover-runner-pages.yml` name is gone; the workflow is now `.github/workflows/deploy-pages.yml`.
- CI now includes a path-scoped changed-pack build check, so touched packs are actually built in PR CI.
- Teletron now has artwork and broad localized descriptions. The generated Pages catalog verified 51 localized descriptions and `https://encorpora.io/assets/teletron-avatar.png`.
- The arm64-only Teletron Rolldown dependency was removed so installs work on GitHub's x64 runners.
- The compiled `corpan/scripts/dev/ipad/ipad-record` Mach-O binary was removed from git; the Swift source remains and the rebuilt binary path is ignored.
- Plus entitlement is now wired into content packs. `ContentPackHost` passes a cached entitlement snapshot at mount, publishes `globalThis.__CORPAN_ENTITLEMENT` / `__CORPAN_PLUS`, and dispatches `corpan:entitlement-changed` when the store changes. Teletron uses that snapshot and event to switch free users to the 20/day quota and Plus users to unlimited without calling Apple/Google IAP per message.

Current status:

- PR is mergeable.
- Latest commit: `9e356246`.
- Checks are all green: `corpan-app · tsc + build`, `web/io · build`, `changed packs · build`, and `terraform · fmt + validate`.
- <0.17 user impact still looks safe through `minAppVersion`, `maxAppVersion`, and `v1Listed: false` catalog gating.

Remaining caveat for later hardening: the Teletron free/Plus quota is client-side enforcement. That is fine for the TestFlight/product validation pass, but a future server-side entitlement/rate-limit check is still the right next step once this becomes revenue-sensitive.

skyl:
## Final sign-off

Reviewed current PR head `9089ddc9` (`world-plaza` -> `main`).

Merge status:
- PR is non-draft and `MERGEABLE`.
- Latest CI is green: `corpan-app · tsc + build`, `web/io · build`, `changed packs · build`, and `terraform · fmt + validate`.
- Local branch is clean and synced with `origin/world-plaza`.
- Targeted local checks passed at current head: `corpan/packs/corpan-city npm run server:typecheck` and `npm run qa:mp-interaction`. The interaction QA covers duplicate player cleanup, invite routing, trusted sender/recipient chat routing, local-cleaned intent relay, peer result routing, trade routing, and Teletron direct country reveal.
- Quick clean-PR scan did not find obvious committed secrets or key files. The compiled `ipad-record` binary is gone.

Release-critical items that previously blocked merge now look addressed:
- Corpan City no longer routes multiplayer interactions through one shared `player-local` identity.
- Free-text peer chat is mediated before crossing devices and failure paths avoid displaying raw peer text.
- Teletron is included in Pages pack publishing and in changed-pack CI.
- Plus entitlement state is bridged into content packs.
- New Tutomaton is gated to app `>=0.17.0`; the legacy `0.3.2` Tutomaton entry remains available for app `0.16.x`.
- Corpan City and Teletron are gated to `>=0.17.0` and `v1Listed: false`, so older app clients should not receive them.

Expected post-merge behavior: after this merges to `main` and the Pages deploy succeeds, the latest Corpan City pack should be live through `catalog-v3.json` for Corpán `0.17.0+` users.

Remaining concerns I would defer to immediate follow-up, not hold this merge for:
1. Deploy/restart the presence server with the latest server hardening if it has not already been deployed.
2. Do a physical iPad/TestFlight QA pass: ASR mic, NPC chat, Teletron invite/leave/rejoin, Corpan City join/profile/proximity, purchase/Plus entitlement display.
3. Corpan City visual/hardening cleanup from the persistent review: clone cached icon canvases before DOM use, fill garment/vessel icon bodies, clamp icon canvas sizes, guard metal lookups, validate saved identity, and gate/remove production debug window hooks. These are real cleanup items, but I do not see them as release blockers for the first 0.17.1 candidate.
4. Teletron free/Plus quota remains client-side enforcement. Fine for this validation pass; server-side quota/entitlement enforcement should follow before this becomes revenue-sensitive.
5. `corpan/packs/stargate-reader/dist/app.js` is a tracked generated bundle and grew from about 4.1 MB to about 6.6 MB in this PR. Not a blocker, but worth cleaning up in the next repo-size pass if generated pack bundles are supposed to stop living in git.

Recommendation: merge. This PR is large and should get one integrated QA pass immediately after merge/deploy, but I do not see a remaining issue that justifies holding `main` any longer.

PR #285, opened 2026-06-09
Teletron 0.1.9 + Corpan City 0.1.7: reliable chat, safe moderation, stable voice, 50-lang chrome, Android assets

skyl:
Ships **Teletron 0.1.9** and **Corpan City 0.1.7**, plus an Android host fix.

## Teletron 0.1.9
- **Reliable chat delivery (the big one).** The accepted-pair link that gates delivery is volatile (lost on a server redeploy, a 24h TTL lapse, or a fresh join after the reconnect window), and the client never re-established it — so `chat-send` was **silently dropped before it ever reached the outbox**. That is why two online people saw nothing, offline messages never arrived, and re-entering a conversation felt dead. Now the server replies `chat-control { link-stale }` instead of dropping; the client queues the message and silently re-invites; an established penpal (durable on-device transcript) is auto-accepted; the queue flushes on confirm. Live delivery, offline→online drain, and leave→rejoin all work. **8 new end-to-end transport tests** against the real `PlazaRoom`. (Durable-across-restart outbox is specced as the next layer in `docs/TRANSPORT_ARCHITECTURE.md`.)
- **Moderation: classify + paraphrase.** Split the overloaded gate into a narrow BLOCK/SAFE classifier + a separate paraphrase. Off-limits messages are obliterated via a corpus remix (raw text never restated); safe messages keep their everyday subject/vocab instead of collapsing to noise. Red-teamed against the real on-device Qwen3-4B (177 cases): **0 classifier leaks, 0% benign over-block, 94% connection kept**; all of the hard innuendo/identity/location/contact vectors BLOCK. No wordlists, no examples in prompts.
- **Stable per-conversation voice** (no more voice jumping between sentences).
- **Notifications, not screen-jacking:** a message from a non-active sender badges the Conversations screen instead of yanking you into the room.
- **No keyboard-pop:** the text field no longer auto-focuses on thread open.
- **Warm copy reframe** away from penpal/private-by-default; "Tidying up…" → "Sending…".
- **50-language chrome** (was English-only), with correct writing systems — Simplified vs Traditional Chinese and Shahmukhi vs Gurmukhi Punjabi kept distinct.

## Corpan City 0.1.7
- Adopts the same two-step mediated-chat moderation (rebuilt from `shared/moderation`).
- Server replies `link-stale` instead of silently dropping an unpaired `chat-send` (additive `ChatControlAction`; in-world City chat ignores unknown actions).

## corpan-app
- **Android pack image/font/audio assets** now load: installed-pack asset URLs were hardcoded `corpan-pack://localhost/…` everywhere, but Tauri serves that scheme at `http://corpan-pack.localhost/…` on Android. URL is now platform-aware; desktop/iOS unchanged.

## Verification
- teletron: tsc ✓, 12 tests ✓, build ✓
- corpan-city: tsc ✓, **716 tests ✓** (incl. 8 new transport e2e), server tsc ✓, build ✓
- corpan-app: tsc ✓, cargo check ✓
- moderation: live red-team battery ✓; i18n script resolution verified at runtime ✓

🤖 Generated with [Claude Code](https://claude.com/claude-code)


skyl:
## 🤖 Codex CLI review

_Automated review via `codex exec` (gpt-5-codex). Posted by the integrator; findings being addressed before merge._

## Summary

The PR is directionally solid, and the Android dual-form pack URL handling looks contained by explicit host matching plus `sanitize_rel`/canonical containment. I found a few concrete issues in the transport self-heal, moderation fallback, and i18n generator paths that should be fixed before merging.

## Blocker

- `corpan/packs/teletron/src/main.ts:1286` / `corpan/packs/teletron/src/main.ts:1880` - A send made while `confirmedPairs` is locally true is not queued before `room.send(MP_MSG.chatSend, input)`. If the server has actually forgotten the pair (TTL lapse/server state reset while the client still believes it is confirmed), Plaza returns `link-stale`, but the client no longer has the prepared `input` to resend, so the user-visible/persisted message is silently lost. The same risk exists after `flushPendingOutbound` deletes the queue before sending (`main.ts:1273`). Suggest retaining the attempted `interactionId`/input until the stale path is ruled out, or have the server include the rejected `interactionId` in `link-stale` so the client can requeue exactly that message; add a test for “confirmed locally, stale on server” and “flush gets link-stale again”.

## Should-fix

- `corpan/packs/teletron/src/main.ts:922` / `corpan/packs/teletron/src/main.ts:1801` - `isEstablishedPenpal()` returns true for any existing conversation, including `lifecycle: "ended"`, so a former partner can re-invite after the user ended the conversation and be silently auto-accepted. A subsequent message then revives the ended thread (`main.ts:1395`) without human consent. That weakens the “stranger/previously revoked link must be human-accepted” model. Suggest auto-accepting only non-ended/non-lapsed links, e.g. require `conversation.lifecycle !== "ended"` and transcript metadata without `lapsedAt`; ended links should fall through to the normal prompt or decline.

- `corpan/packs/shared/moderation/index.ts:789` - On the SAFE path, if `relay.paraphrase` returns empty/protocol junk, `const seed = paraphrase || raw` feeds the raw user message into `relay.regenerate`. That contradicts the new firewall guarantee and can leak raw UGC to the relay text if the regenerator echoes a benign sentence or a classifier false-negative. Suggest treating paraphrase failure as a replacement/fallback path using corpus seeds, not `raw`; add a unit test where `relay.classify=SAFE`, `relay.paraphrase=""`, and `relay.regenerate` tries to echo the input.

- `corpan/packs/teletron/tools/gen_i18n.py:184` - Script-variant preservation only applies to `--from-json`; the normal API generation path still collapses `zh-Hans`/`zh-Hant` and `pa-Arab`/`pa-Guru` to `code.split("-")[0]` and skips the second variant once the base exists. The next full regeneration without `--from-json` can overwrite the checked-in variant keys with a single base locale. Suggest using `loc_key()` and `BASE_ALIAS` in the API path too, matching `from_json()`.

## Nit

- No nit findings. The pack URL parsing change does not appear to introduce path traversal or SSRF: the new local URL forms are recognized before network fetch, and both text/bytes reads still sanitize segments and canonicalize under the pack root.

skyl:
## ✅ Codex findings addressed (commit d4f20eaf)

- **Blocker — message loss on confirmed-locally/stale-on-server:** confirmed-path sends are now retained in a bounded in-flight ring; the server echoes the rejected `interactionId` in `link-stale`, and the client requeues that exact message. New server test asserts the echoed id.
- **Should-fix — ended-link auto-accept:** `isEstablishedPenpal()` now returns false for an `ended` live link or a `lapsed` transcript, so a closed conversation can't be silently revived without a fresh human accept.
- **Should-fix — moderation firewall (`seed = paraphrase || raw`):** a failed paraphrase no longer falls back to the raw message; it routes to the safe corpus seed. New unit test (`classify=SAFE`, `paraphrase=""`) proves no raw leak + `state: replaced`.
- **Should-fix — i18n API path:** `loc_key()`/`BASE_ALIAS` now apply on the API generation path too, so a full regen can't clobber the `zh-Hant`/`pa-Arab` script variants.

Gate after fixes: **717 corpan-city + 12 teletron tests green**, all tsc clean, both packs rebuilt.

PR #302, opened 2026-06-18
Juice Squeeze — ground-up rebuild, shipped as the juice_squeeze upgrade (v0.1.6)

Umanistan:
A ground-up rebuild of Juice Squeeze (React + @dnd-kit + Canvas-2D liquid, layered Web Audio, native haptics, basket→coins meta-loop). Per review, it now ships as the **in-place `juice_squeeze` upgrade** — not a separate pack: it lives in `corpan/packs/juice-squeeze/`, keeps id **`juice_squeeze`**, bumps **0.1.5 → 0.1.6**, and keeps the persist key (`juice-squeeze-game-state`) so existing players' bottle collections carry over. The existing deploy-pages/CI pipeline already builds that dir, so the catalog output stays aligned. **Pack-side only — no corpan-app/host changes.**

## Try on a device (no local build)
corpan **0.18.0+** → Settings → tap "Corpan" 7× → Packs → paste:
`https://github.com/Umanistan/encorpora/releases/download/juice-squeeze-v0.1.6-preview/juice-squeeze.zip`

## Review items addressed
- **Level-complete review emptied by a basket on a level boundary** — fixed: the completed bottle's phrases are captured into `LevelCompleteInfo` and the modal renders from that snapshot, not the mutable shelf.
- **EN→EN** — hardened: `pickByLang` no longer silently falls back to a wrong language; `loadUtterance` rejects entries whose block/target text are identical; `pickLanguagePair` dedupes and never returns an identical pair when avoidable.
- **Android last-word TTS clip** — the per-word voice is skipped on the tap that places the final word (the win-sentence TTS reads it moments later); snap+haptic still fire.
- **Mobile header / vertical space** — header breathing room clear of the status area + a middle-band vertical rhythm (auto-margins, no fixed heights; the long-phrase auto-fit is preserved).
- **Audio not loading in the installed zip** — root cause was a hand-built asset URL that mis-resolved the `corpan-pack://` scheme. Now each WAV is ES-imported (vite inlines as a data URL) and played via Web Audio with an HTMLAudio fallback — hover-runner's proven pattern. Works on every host version (incl. 0.18.0), iOS/Android/desktop, offline. Verified end-to-end in the installed zip.
- **Auto-advance** — intentional design (kept); corrected the stale "NO auto-advance" header comment.
- **Eyeball reveal silent** — intentional: 👁 reveals (see), 🎧 speaks (hear). Kept silent by design.

tsc clean, 5/5 tests, build + pack verified.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


corpan/packs/juice-squeeze/src/hooks/useGameLogic.ts
```diff
@@ -0,0 +1,509 @@
...
+    const tVoice = window.setTimeout(() => {
+      if (mounted.current) speak(blockLang, sentence)
+    }, VOICE_DELAY)
+    winTimers.current.push(tVoice)
+
+    updateDebug({ hasWon: true })
+
+    // ---- NORMAL win: advance after the voice (scaled to sentence length) -----
+    if (!willComplete) {
+      advanceTimer.current = window.setTimeout(() => {
```
skyl, line 325: The file header says this keeps shipped parity and should not auto-advance after a win, but this timer replaces the solved phrase as soon as the voice/buffer delay expires. The shipped pack leaves the completed sentence on screen so the player can inspect/replay it and advance manually. If auto-advance is the intended new behavior, the parity note and tests need to change; otherwise this should stay on the completed phrase until Next/swipe/give-up.

corpan/packs/juice-squeeze/src/hooks/useGameLogic.ts
```diff
@@ -0,0 +1,509 @@
...
+    // mint a coin FIRST (its own beat), then show the level modal / advance.
+    const tDock = window.setTimeout(() => {
+      if (!mounted.current) return
+      const shelf = useGameStore.getState().bottleProgress.bottleCollection
+      let basketDelay = 0
+      if (shelf.length >= BASKET_SIZE) {
+        basketDelay = BASKET_CARRY_MS
+        basketCleanup.current = launchBasketCarry({
+          // Jars cloned → clear the real shelf so they don't double up.
+          onStart: () => useGameStore.getState().removeBasketJars(BASKET_SIZE),
```
skyl, line 375: Clearing the shelf here breaks the level-complete review when the level finishes on a basket boundary. At exactly 6 collected jars, `removeBasketJars(6)` removes the just-completed bottle before `setLevelComplete(...)`; `LevelCompleteModal` then reads `bottleCollection[bottles.length - 1]` and shows `noPhrases`. Capture/pass the completed bottle's phrases before clearing the basket, or render review from level-complete state instead of the mutable shelf.

skyl (dismissed):
I found two behavior regressions that should be fixed before merge.

Checks run:
- npm ci
- npm run typecheck
- npm run test:run
- npm run build
- npm run pack

skyl:
Live Android testing follow-up.

Device screenshot captured via `adb` from `SM_S938U` / `R5CY847ZCVN` (`1080x2340`, density `450`):

Android screenshot: cramped Juice Squeeze Fire mobile layout

Additional feedback from live testing:

- Android TTS cuts off the last placed word. When the final word is placed, the sentence/win TTS starts and appears to interrupt the word-level TTS. iOS can overlap multiple utterances, but Android seems to replace the previous utterance, so the last tap feels clipped. If the host has a queued TTS API, this path should use it; otherwise the win sentence playback needs to wait long enough for the final word utterance to finish instead of overrunning it.
- The mobile header needs more space and safer breathing room. The score/coin/jar cluster, current-fruit control, and exit button feel crowded near the system status area.
- The mobile layout is not using vertical space well. The controls and word bank might need to be anchored/fixed near the bottom, with the target/sentence area taking a more deliberate middle band. In the screenshot, the interactive bank/control area feels cramped while a large lower portion of the screen is unused; this needs a more premium spacing pass.

skyl:
Additional live-testing / ship-readiness feedback:

- The pack can show EN -> EN: the prompt is English and the placed blocks are also English. That should not happen. Likely causes to harden:
  - `pickLanguagePair()` currently permits same-language pairs for empty/single-language stacks and duplicate language arrays (`["en", "en"]`).
  - `pickByLang()` falls back to `translations[0]` when the requested language is missing, so a requested target language can silently become the same language as the block text. The loader should reject entries that do not have distinct usable target/block translations for this game mode, and the language picker should avoid identical target/block pairs.
- The eyeball/show-answer button should speak the answer TTS when the modal opens. Current code documents and implements the eye reveal as silent, but the expected behavior from testing is reveal + speak.
- The zip used in testing appears to have no sounds loading. The local `pack.mjs` includes `manifest.json + dist/`, and Vite should copy `public/audio` into `dist/audio`, but the actual distributed/catalog artifact needs to be verified end-to-end. Please double-check the asset packaging standard against the existing pack pipeline and confirm the release zip contains and resolves `dist/audio/*.wav` from the installed pack origin.
- This should probably not ship as a new `juice_squeeze_fire` pack ID. It should replace/upgrade the existing `juice_squeeze` pack as a minor version bump of regular Juice Squeeze, with the manifest ID kept as `juice_squeeze` and the version bumped correctly from the current shipped `juice-squeeze` version.
- Before merge, hook the real release build into the existing catalog/GitHub Pages pipeline. The current Pages workflow copies/packages `corpan/packs/juice-squeeze`, but this PR adds `corpan/packs/juice-squeeze-fire`; as a new pack ID it is not on the normal catalog path. If this is intended as the Juice Squeeze upgrade, the implementation should land in the existing `corpan/packs/juice-squeeze` package and keep the catalog/GH Actions/Pages output aligned before merging.

Umanistan:
Thanks for the thorough review, Skylar — all addressed:

- **Level-complete basket bug (your #2):** genuine bug, fixed. The just-completed bottle's phrases are now captured into `LevelCompleteInfo` at level-complete time and the modal renders from that snapshot, so a basket clearing the shelf on a boundary can't empty the review.
- **Auto-advance (your #1):** intentional (Ian confirmed) — it's the fast-paced flow, and players can review via prev/swipe. I corrected the stale "NO auto-advance (parity)" header comment to document it; no behavior change.
- **EN→EN:** hardened all three paths — `pickByLang` is strict (no silent fallback), `loadUtterance` skips identical block/target text, and `pickLanguagePair` dedupes + avoids identical pairs (single-language stacks pair with English). Test updated.
- **Android last-word TTS clip:** the per-word voice is now skipped on the final-word tap; the win-sentence TTS covers it, snap+haptic still fire.
- **Mobile header + vertical layout:** header breathing room + a middle-band rhythm via auto-margins (no fixed heights — the long-phrase auto-fit/scroll/collapse is intact).
- **Audio not loading in the zip:** root cause was a hand-built URL mis-resolving `corpan-pack://`. Switched to ES-imported WAVs (vite data URLs) + Web Audio with an HTMLAudio fallback (hover-runner's pattern) — verified working in the installed pack on 0.18.0 and 0.19.0.
- **Eyeball reveal silent:** intentional — 👁 = see (reveal), 🎧 = hear (speak). Keeping it silent by design.
- **Ship as the juice_squeeze upgrade:** done — moved into `corpan/packs/juice-squeeze/`, id `juice_squeeze`, v0.1.5→0.1.6, persist key unchanged so collections carry over; the existing deploy-pages/CI pipeline builds that dir so the catalog stays aligned. No host changes.

PR #539, opened 2026-07-26
ci: add the native/Rust gate (fmt ratchet, compile, clippy, patch integrity)

skyl:
## What

Add a `native` job to `ci.yml` that runs rustfmt, `cargo clippy --locked --all-targets`
and a Cargo `[patch]` integrity assertion over the Rust tree, and wire it into
`ci-gate`'s `needs`.

## Why

Nothing required in this repo ran `cargo`, `rustfmt`, `clippy` or `rustup`. Every
Rust change in both shipping apps merged uncompiled, and the first signal was a
release build at store-submission time. `ios-native.yml` compiles plugin Swift but
is advisory, has no `merge_group` trigger, and carries a top-level `paths:` filter —
which is exactly why it can never be promoted to a required context: it would not
report on unrelated PRs and would deadlock the merge queue permanently.

The hook was already built and unconsumed: `changes` has been declaring
`set_output native '^(native/|.*/src-tauri/|corpan/plugins/)'` with no job reading it.

## Blast radius

**Areas touched:** CI

**Why this cannot deadlock the merge queue.** The `native` job is path-gated, so it
is *not* a required status context — it reports into `ci-gate`, which is already
required and always reports (`if: always()`). Adding a `needs` entry to an
always-reporting aggregate adds coverage without adding a context branch protection
is waiting on. `ci-gate` already treats `skipped` as a pass, so a docs-only PR that
skips `native` still gets a green gate. Zero queue risk by construction.

- [x] Every touched path is covered by a `ci.yml` area filter. This PR *tightens*
      the `uncovered` warning: `src-tauri/` is no longer blanket-ungated, and the
      four trees this job does not compile are named explicitly in `ungated`
      (`corpus-reader/src-tauri/`, `panko/pako/src-tauri/`,
      `homeschool-offline/app/src-tauri/`, `corpan/plugins/tauri-plugin-iap/examples/`).
- [x] **Pack back-compat.** N/A — no pack touched.
- [x] **Native integrity.** No Rust source changed. This PR is what starts
      enforcing that checkbox: it asserts no `[[patch.unused]]` in any tracked
      lock, that `cargo metadata --locked` emits no "was not used in the crate
      graph" on **stderr** for every manifest declaring a `[patch]` of any shape
      (path, git or version), and that each *path* patch resolves to its vendored
      directory. That is the exact failure mode that once reverted the vendored
      `ndk-context` fork.
- [x] **Strings.** N/A — no user-visible string.
- [x] **Curriculum.** N/A.
- [x] **Secrets.** None.

## What it does, and what it deliberately does not

**Compiles** `cargo clippy --locked --all-targets` on both shipping `src-tauri`
trees. Clippy rather than `cargo check` because clippy *is* a check plus lints:
one compile, both signals. A compile error fails the job (cargo's exit code, carried
by `pipefail`). Lint warnings are de-duplicated, counted, rendered, and held to a
**bidirectional budget** — `CLIPPY_BUDGET=16`, the measured debt — so the first PR
to touch Rust is not blocked by pre-existing debt, but the count cannot drift up
unnoticed, and driving it down without lowering the budget fails with the number to
write.

Corpán's `src-tauri` declares ten plugin path deps; the eleventh crate,
`corpan-asr-contract`, is reached transitively via `tauri-plugin-asr-native`. So
compiling the app compiles all eleven — including `tauri-plugin-stt`, which
`ios-native.yml` skips. Verified against the resolved graph:

```
$ cargo metadata --locked --manifest-path corpan/corpan-app/src-tauri/Cargo.toml
  # packages under corpan/plugins/ →
  corpan-asr-contract, tauri-plugin-asr-native, tauri-plugin-audio-keepalive,
  tauri-plugin-corpan-llm, tauri-plugin-game-packs, tauri-plugin-haptics,
  tauri-plugin-iap, tauri-plugin-radio-stream, tauri-plugin-stt,
  tauri-plugin-subscriptions, tauri-plugin-tts        # count = 11
```

**rustfmt as a bidirectional ratchet.** Fourteen of the eighteen tracked crates
carry formatting debt, so a blanket `--check` would fail on day one and get deleted.
Every crate *not* on the `DEBT` list is enforced, including every crate added from
here on. Formatting a debt crate without deleting its line also fails, with the
exact line to delete — that is what keeps the ratchet from rusting open. The
toolchain is deliberately unpinned; the split was measured as identical under
rustfmt 1.8.0 (rustc 1.93.1, local) and rustfmt 1.9.0-stable (rustc 1.97.1, what
the runner shipped), with identical per-crate diff counts.

**Not covered, on purpose** (named in the workflow comments rather than papered over):

- `#[cfg(mobile)]` code. A Linux host build takes the `desktop` cfg, so each
  plugin's `mobile.rs` and Corpán's Android JNI / crash-breadcrumb blocks are not
  compiled. iOS Rust+Swift is advisory-only via `ios-native.yml`; the **Android
  Rust glue is covered by nothing**. Closing that needs an NDK on the runner.
- Clippy *lints* only the two app crates. Path deps outside the workspace directory
  do not get the clippy wrapper, so the plugins are compile-checked, not linted.
- `corpus-reader`, `panko/pako` and `homeschool-offline` are **not compiled**.
  `corpus-reader` and `panko` are dormant; `panko` additionally resolves a plugin
  from a git *branch*; `homeschool-offline` is the only one of the three with no
  tracked `Cargo.lock`, so it alone could not sit behind `--locked` today.
  (`git ls-files '*Cargo.lock'` returns four: corpan, corpus-reader, dynawalla,
  panko.) The rustfmt ratchet **walks** all of them, but `corpus-reader` and both
  `homeschool-offline` crates are on the `DEBT` list, so they are reported as
  `::warning::` and cannot fail the job — walked, not gated. `panko/pako/src-tauri`
  and the iap example are off the list and are enforced. The `uncovered` step names
  all four so the warning keeps pointing at the real remaining hole.

## Two second-order costs, stated up front

- `native` self-triggers on `.github/workflows/ci.yml`, like every other filter that
  drives a job, so every future CI-config PR pays a ~4–7 min Rust build in both the
  PR and the `merge_group` run. That is deliberate: a workflow edit that breaks the
  native gate has to be caught by the native gate.
- The first `merge_group` entry after this merges runs cold (~7 min) because no
  `main`-scoped rust-cache exists yet — inside both the job's 30 min timeout and the
  queue's 60 min `check_response_timeout_minutes`.

## Proof

See the checks on this PR, and the comments below for measured timings, the
aggregation invariant, the both-directions filter proof, and the review round that
closed the `[patch]` fail-open and added the clippy budget.

### DIFF
--- a/.github/scripts/clippy_report.py
+++ b/.github/scripts/clippy_report.py
@@ -27,14 +27,27 @@ import sys
 from typing import IO
 
 
-def diagnostic_key(message: dict) -> tuple:
-    """Identity of a diagnostic, stable across the targets that re-emit it."""
+def diagnostic_key(record: dict) -> tuple:
+    """Identity of a diagnostic, stable across the targets that re-emit it.
+
+    Takes the whole cargo envelope, not just `message`, because `file_name` in
+    the span is relative to the PACKAGE root: two crates each warning at
+    `src/lib.rs:42:5` with the same lint and text are indistinguishable without
+    `package_id`, and the second would be both uncounted and unprinted. The job
+    compiles eleven plugin crates as path deps (cargo does not `--cap-lints
+    allow` a path source), so that collision is reachable.
+
+    `--all-targets` re-emits a lib diagnostic for the lib-test target under the
+    SAME `package_id`, so cross-target de-duplication is unaffected.
+    """
+    message = record.get("message") or {}
     code = (message.get("code") or {}).get("code") or ""
     spans = message.get("spans") or []
     primary = next((s for s in spans if s.get("is_primary")), None)
     if primary is None:
         primary = spans[0] if spans else {}
     return (
+        record.get("package_id") or "",
         code,
         primary.get("file_name", ""),
         primary.get("line_start", 0),
@@ -64,7 +77,7 @@ def process(stream: IO[str], out: IO[str]) -> tuple[int, int]:
         if level not in ("warning", "error"):
             continue
 
-        key = diagnostic_key(message)
+        key = diagnostic_key(record)
         (errors if level == "error" else warnings).add(key)
 
         if key not in printed:
--- a/.github/scripts/clippy_report.py
+++ b/.github/scripts/clippy_report.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python3
+"""Render `cargo --message-format=json` diagnostics and count the unique ones.
+
+Why not parse cargo's own human summary: it prints
+
+    warning: `corpan` (lib) generated 2 warnings (1 duplicate)
+    warning: `corpan` (lib test) generated 2 warnings (1 duplicate)
+
+for `--all-targets`, tagging BOTH lines as containing duplicates. Summing them
+double-counts; dropping every line that says "duplicate" counts zero. Neither is
+the number a human wants. So: read the structured stream, de-duplicate on the
+diagnostic's own identity (lint code + span + text), and print that.
+
+Usage:
+    cargo clippy --message-format=json ... | clippy_report.py --count-file F
+
+stdout of cargo is JSON; its stderr (progress, the final error summary) is left
+alone and flows straight to the job log. This reads stdin to EOF — it must never
+exit early, or cargo takes SIGPIPE and the pipeline returns 141.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+from typing import IO
+
+
+def diagnostic_key(message: dict) -> tuple:
+    """Identity of a diagnostic, stable across the targets that re-emit it."""
+    code = (message.get("code") or {}).get("code") or ""
+    spans = message.get("spans") or []
+    primary = next((s for s in spans if s.get("is_primary")), None)
+    if primary is None:
+        primary = spans[0] if spans else {}
+    return (
+        code,
+        primary.get("file_name", ""),
+        primary.get("line_start", 0),
+        primary.get("column_start", 0),
+        message.get("message", ""),
+    )
+
+
+def process(stream: IO[str], out: IO[str]) -> tuple[int, int]:
+    """Print rendered diagnostics; return (unique warnings, unique errors)."""
+    warnings: set[tuple] = set()
+    errors: set[tuple] = set()
+    printed: set[tuple] = set()
+
+    for line in stream:
+        line = line.strip()
+        if not line or not line.startswith("{"):
+            continue
+        try:
+            record = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+        if record.get("reason") != "compiler-message":
+            continue
+        message = record.get("message") or {}
+        level = message.get("level")
+        if level not in ("warning", "error"):
+            continue
+
+        key = diagnostic_key(message)
+        (errors if level == "error" else warnings).add(key)
+
+        if key not in printed:
+            printed.add(key)
+            rendered = message.get("rendered")
+            if rendered:
+                out.write(rendered if rendered.endswith("\n") else rendered + "\n")
+
+    return len(warnings), len(errors)
+
+
+def main(argv: list[str]) -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--count-file",
+        help="File to append the unique warning count to (one integer per line).",
+    )
+    parser.add_argument(
+        "--label", default="", help="Prefix for the summary line, e.g. the crate path."
+    )
+    args = parser.parse_args(argv)
+
+    warnings, errors = process(sys.stdin, sys.stdout)
+
+    label = f"{args.label}: " if args.label else ""
+    print(f"{label}{warnings} unique clippy warning(s), {errors} unique error(s)")
+
+    if args.count_file:
+        with open(args.count_file, "a", encoding="utf-8") as fh:
+            fh.write(f"{warnings}\n")
+
+    # Always 0: the compile verdict is cargo's exit code, which `pipefail`
+    # propagates. Failing here too would report the same failure twice and
+    # would make a lint warning look like a build break.
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main(sys.argv[1:]))
--- a/.github/scripts/patch_integrity.py
+++ b/.github/scripts/patch_integrity.py
@@ -1,0 +1,219 @@
+#!/usr/bin/env python3
+"""Assert every `[patch.*]` in the repo is actually applied.
+
+A Cargo patch that stops matching does not fail a build. Cargo emits
+
+    warning: Patch `foo v1.2.3 (/path)` was not used in the crate graph.
+
+on **stderr**, exits 0, and quietly resolves the upstream crate instead. This
+repo has already lost a vendored crash fix that way: the patched crate was
+silently dropped, the app compiled, tested and clippied clean, and the
+regression only surfaced on a device.
+
+Three assertions, in increasing strength:
+
+1. No tracked `Cargo.lock` contains a `[[patch.unused]]` stanza. Cargo writes
+   that section when a declared patch does not apply, so a lock is a durable,
+   grep-able record of the failure.
+2. `cargo metadata` for each patching manifest emits no "was not used in the
+   crate graph" warning. This is the load-bearing one — `cargo metadata` exits
+   0 either way, so the *only* signal is the stderr text.
+3. Every patched package resolves in the graph to the vendored path the patch
+   declares, not to a registry copy.
+
+Run with the manifests to inspect, or with no arguments to inspect every
+tracked manifest that declares a patch.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import re
+import subprocess
+import sys
+import tomllib
+from pathlib import Path
+
+# Cargo's wording for a patch that resolved to nothing. Matched as a substring
+# because the surrounding text (crate name, version, source) varies.
+UNUSED_PATCH_MARKER = "was not used in the crate graph"
+
+# Third-party trees we neither format nor own. A vendored fork's own manifest
+# is exempt from inspection; the patch that POINTS at it is not.
+EXCLUDED_PATH_PARTS = ("vendor", "node_modules", "target")
+
+
+def has_unused_patch_stanza(lock_text: str) -> bool:
+    """True when a Cargo.lock records a patch that did not apply."""
+    return re.search(r"^\[\[patch\.unused\]\]", lock_text, re.MULTILINE) is not None
+
+
+def unused_patch_warnings(stderr: str) -> list[str]:
+    """The `cargo` stderr lines reporting a patch that matched nothing."""
+    return [line.strip() for line in stderr.splitlines() if UNUSED_PATCH_MARKER in line]
+
+
+def patch_entries(manifest: Path) -> list[tuple[str, str, Path]]:
+    """(registry, crate, resolved path) for every path-based patch in a manifest.
+
+    Patches that are not path-based (git, version-only) are returned with a
+    path of ``None`` by the caller's contract — none exist in this repo today,
+    so they are skipped rather than guessed at.
+    """
+    data = tomllib.loads(manifest.read_text(encoding="utf-8"))
+    out: list[tuple[str, str, Path]] = []
+    for registry, crates in (data.get("patch") or {}).items():
+        for crate, spec in crates.items():
+            path = spec.get("path") if isinstance(spec, dict) else None
+            if path is None:
+                continue
+            out.append((registry, crate, (manifest.parent / path).resolve()))
+    return out
+
+
+def tracked_files(pattern: str, repo_root: Path) -> list[Path]:
+    out = subprocess.run(
+        ["git", "ls-files", "-z", pattern],
+        cwd=repo_root,
+        capture_output=True,
+        text=True,
+        check=True,
+    ).stdout
+    paths = [repo_root / p for p in out.split("\0") if p]
+    return [
+        p for p in paths if not any(part in EXCLUDED_PATH_PARTS for part in p.parts)
+    ]
+
+
+def cargo_metadata(manifest: Path) -> tuple[dict, str]:
+    """Full (dependency-resolving) metadata plus cargo's stderr.
+
+    `--no-deps` would skip resolution entirely, which is exactly the step that
+    decides whether a patch applies — so it must not be used here.
+    """
+    proc = subprocess.run(
+        [
+            "cargo",
+            "metadata",
+            "--locked",
+            "--format-version",
+            "1",
+            "--manifest-path",
+            str(manifest),
+        ],
+        capture_output=True,
+        text=True,
+    )
+    if proc.returncode != 0:
+        raise SystemExit(
+            f"cargo metadata failed for {manifest} (exit {proc.returncode}):\n"
+            f"{proc.stderr}"
+        )
+    return json.loads(proc.stdout), proc.stderr
+
+
+def check_locks(repo_root: Path) -> list[str]:
+    failures = []
+    locks = tracked_files("*Cargo.lock", repo_root)
+    for lock in locks:
+        rel = lock.relative_to(repo_root)
+        if has_unused_patch_stanza(lock.read_text(encoding="utf-8")):
+            failures.append(
+                f"{rel}: contains a [[patch.unused]] stanza — a declared "
+                f"[patch] no longer matches anything in the graph."
+            )
+        else:
+            print(f"  ok  {rel}: no [[patch.unused]]")
+    if not locks:
+        failures.append("no tracked Cargo.lock found — is this the repo root?")
+    return failures
+
+
+def check_manifest(manifest: Path, repo_root: Path) -> list[str]:
+    rel = manifest.relative_to(repo_root)
+    failures = []
+    entries = patch_entries(manifest)
+    if not entries:
+        return failures
+
+    metadata, stderr = cargo_metadata(manifest)
+
+    for line in unused_patch_warnings(stderr):
+        failures.append(f"{rel}: cargo reported an unapplied patch: {line}")
+
+    by_name: dict[str, list[dict]] = {}
+    for pkg in metadata.get("packages", []):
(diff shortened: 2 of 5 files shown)


skyl:
## Verification

### 1. `ci-gate` aggregation invariant (PyYAML)

```
$ python - <<'PY'
import yaml
d = yaml.safe_load(open(".github/workflows/ci.yml"))
jobs = set(d["jobs"]); needs = set(d["jobs"]["ci-gate"]["needs"])
print("not aggregated :", sorted(jobs - {"ci-gate", "changes"} - needs))
print("needs w/o a job:", sorted(needs - jobs))
PY
top-level jobs      : ['changed-packs', 'changes', 'ci-gate', 'ci-scripts', 'corpan-app', 'dja', 'dynawalla-android-gen', 'dynawalla-app', 'dynawalla-curriculum', 'dynawalla-engine', 'lambda', 'native', 'pack-catalog', 'terraform', 'web-io']
ci-gate.needs       : ['changed-packs', 'changes', 'ci-scripts', 'corpan-app', 'dja', 'dynawalla-android-gen', 'dynawalla-app', 'dynawalla-curriculum', 'dynawalla-engine', 'lambda', 'native', 'pack-catalog', 'terraform', 'web-io']
not aggregated      : []
needs w/o a job     : []
```

The "not aggregated" list is empty: `native` feeds `ci-gate`, and no other job has
been left out. `needs w/o a job` is empty too, so nothing in `needs` is a typo that
would silently never report.

### 2. Both directions of the path filter

Driven through the **verbatim** `changes` → `Detect changed areas` script, extracted
from `ci.yml` with PyYAML and run against two real single-file commits.

```
$ python -c "import yaml,sys; d=yaml.safe_load(open('.github/workflows/ci.yml'));
             s=[x for x in d['jobs']['changes']['steps'] if x.get('id')=='filter'][0];
             open('filter_step.sh','w').write(s['run'])"

===== CASE A: docs-only PR  (fb6f5e6bf..8ccb51e6a) =====
Changed files:
  README.md
--- resolved outputs ---
  ci_scripts=false
  corpan_app=false
  dja=false
  dynawalla_android_gen=false
  dynawalla_app=false
  dynawalla_curriculum=false
  dynawalla_engine=false
  lambda=false
  native=false          <-- native job SKIPS
  packs=false
  shared_ts=false
  terraform=false
  web=false

===== CASE B: Rust-only PR  (8ccb51e6a..0f9eb1cf1) =====
Changed files:
  corpan/plugins/tauri-plugin-tts/src/lib.rs
--- resolved outputs ---
  ci_scripts=false
  corpan_app=false
  dja=false
  dynawalla_android_gen=false
  dynawalla_app=false
  dynawalla_curriculum=false
  dynawalla_engine=false
  lambda=false
  native=true           <-- native job RUNS, and it is the only one
  packs=false
  shared_ts=false
  terraform=false
  web=false
```

Case A is the docs-only PR: every filter resolves `false`, so `native` is skipped —
and so is every other area job.
Case B is the Rust-only PR: `native` is the **only** filter that flips, so a plugin
edit runs exactly this job and nothing else.

### 3. `ci-gate` passes when its needs are skipped

Not a claim about new code — `ci-gate`'s aggregation is untouched, and it treats
`skipped` as a pass. Most recent CI run on `main`
([30201827033](https://github.com/corpora-inc/encorpora/actions/runs/30201827033),
commit `9bc35eed9`): **11 of 13 area jobs skipped, `ci-gate` succeeded.**

```
skipped  changed packs · test + build
skipped  ci-scripts · tests
skipped  dja · tests
skipped  dynawalla · gen/android template drift
skipped  dynawalla-app · lint + tsc + test + build
skipped  dynawalla-curriculum · tests + gates
skipped  dynawalla-engine · tests
skipped  lambda · tests
skipped  pack-catalog · integrity
skipped  terraform · fmt + validate
skipped  web/io · build
success  changes
success  ci-gate
success  corpan-app · tsc + build
```

A pre-merge empirical skip of `native` itself is not obtainable: `changes` resolves
its range as `base.sha..head.sha`, and for any PR that *adds* this job the range
necessarily contains `ci.yml`. §2 runs the real filter code against a real docs-only
range instead; §3 is the real evidence for what `ci-gate` does with a skip.

### 4. Real CI timings

Six real runs on this PR. Job is `native · fmt + compile + clippy`, `ubuntu-latest`.

| run | cache | job total | rustfmt | patch integrity | Corpán compile+clippy | Dynawalla |
|---|---|---|---|---|---|---|
| [30202470247](https://github.com/corpora-inc/encorpora/actions/runs/30202470247) | stone cold | **7m18s** | 2s | 4s | 265s | 67s |
| [30202775086](https://github.com/corpora-inc/encorpora/actions/runs/30202775086) | full match | **4m15s** | 5s | 1s | 171s | 2s |
| [30203996674](https://github.com/corpora-inc/encorpora/actions/runs/30203996674) | full match | **4m09s** | 5s | 1s | 174s | 1s |
| [30204226665](https://github.com/corpora-inc/encorpora/actions/runs/30204226665) (head) | full match | **4m14s** | 5s | 1s | ~175s | 1s |

Warm step breakdown of the final run:

```
13:25:50  checkout                       13s
13:26:03  git lfs pull (icons only)       1s
13:26:04  toolchain (rustup component)    2s
13:26:06  apt-get (webkit2gtk, cmake…)   33s
13:26:39  rust-cache restore             15s
13:26:54  rustfmt ratchet (18 crates)     5s
13:26:59  cargo [patch] integrity         1s
13:27:00  cargo clippy × 2              175s
13:29:57  job end
```

**Disk.** Not a constraint. `/` goes 59G → 62G used of 145G on the cold run and
does not move on warm runs: ~3 GB for both target trees. `cargo clippy` never
links, so the ~148 GB local Corpán tree (release artifacts, iOS/Android targets,
llama binaries) has no CI analogue.

**Where the warm 175s goes, and what does not fix it.** On a *full cache match*
cargo still rebuilds every PATH-sourced crate — the vendored `llama-cpp-sys-2`
(cmake + bindgen over llama.cpp, the biggest single item), all eleven plugins,
and the app crate. rust-cache strips those from the target dir before saving.

`cache-workspace-crates: true` was tried and measured across two more runs
([30203187150](https://github.com/corpora-inc/encorpora/actions/runs/30203187150),
[30203761837](https://github.com/corpora-inc/encorpora/actions/runs/30203761837)):
**156s vs 165s — inside the noise**, because the plugins live outside the
workspace root and are not "workspace crates" by that flag's definition. It is
therefore **not shipped**; the measurement and the real lever (sccache, which can
also front the C++ via `CMAKE_C_COMPILER_LAUNCHER`) are recorded as comments
instead. Shipping a knob whose comment claims a benefit it does not deliver is
worse than not shipping it.

Two rust-cache traps found the expensive way and written down in the workflow:
it does **not** re-save on a full key match (so a caching change looks like a
no-op against an existing cache), and a `pull_request` cache is scoped to that
PR — only the `push: main` run populates a cache every PR can read.

`timeout-minutes` is **30**: headroom over the 7m18s cold measurement, rather
than 45 minutes of queue time burnt before anyone notices a hang.

### 5. Lint / format debt this measured

**Compile: clean.** 0 errors on both apps. This is the first time either
`src-tauri` tree has been compiled by CI.

**Clippy: 16 unique warnings, all in `corpan`; Dynawalla is at 0.**
De-duplicated across the lib and lib-test targets — cargo's own summary reports
`generated N warnings (M duplicates)` on *both* lines, so summing it double-counts
and dropping "duplicate" lines counts zero. `.github/scripts/clippy_report.py`
reads the JSON stream and keys on lint code + span instead.

```
4  using `map_err` over `inspect_err`
2  very complex type used. Consider factoring parts into `type` definitions
2  unnecessary use of `clone` to create a slice from a reference
2  this `map_or` can be simplified
1  unused imports: IMG_SUBDIR, TMP_SUBDIR, img_rel_path, validate_cache_rel
1  this import is redundant
1  this function has too many arguments (9/7)
1  this function has too many arguments (10/7)
1  field `size` is never read
1  clamp-like pattern without using clamp function
```

**What `-D warnings` would take.** Much less than expected: one Rust-only PR.
Twelve are mechanical rewrites, two are `#[allow(clippy::too_many_arguments)]` or
a struct, two are dead code to delete. Then append `-- -D warnings` to the clippy
invocation. Out of scope here — this PR's file scope is `ci.yml` plus its scripts,
and landing `-D warnings` today would fail the first PR that touches Rust.

**rustfmt: 14 of 18 crates carry debt, 208 diff hunks total.** Enforced today on
`tauri-plugin-haptics`, `iap-demo`, `dynawalla/dynawalla-app/src-tauri`,
`panko/pako/src-tauri` — and on every crate added from now on. Reported for:

```
corpan/corpan-app/src-tauri                                    31
homeschool-offline/app/src-tauri                              126
corpus-reader/src-tauri                                        17
corpan/plugins/tauri-plugin-game-packs                          6
corpan/plugins/{tauri-plugin-iap,tauri-plugin-audio-keepalive}   4 each
corpan/plugins/{tauri-plugin-asr-native,-corpan-llm,-stt,-subscriptions}  3 each
corpan/plugins/{tauri-plugin-tts, …ios-share}                   2 each
corpan/plugins/{corpan-asr-contract,tauri-plugin-radio-stream}  1 each
```

The C++ warnings llama.cpp's build script emits (`-Wunused-function` in
`jinja/lexer.h` etc.) are cargo stderr passthrough and are correctly **not**
counted — the 16 above are exactly the Rust diagnostics.

### 6. The `[patch]` assertion actually fails on a broken patch

Proven against a throwaway crate with a vendored fork whose version no longer
satisfies the requirement — the exact silent-drop mode. All three assertions fire:

```
::error::Cargo.lock: contains a [[patch.unused]] stanza — a declared [patch] no longer matches anything in the graph.
::error::Cargo.toml: cargo reported an unapplied patch: warning: patch `libc v0.1.0 (…/vendor/libc)` was not used in the crate graph
::error::Cargo.toml: [patch.crates-io] `libc` should resolve to …/vendor/libc/Cargo.toml but the graph has …/registry/src/index.crates.io-…/libc-0.2.189/Cargo.toml.
EXIT=1
```

`cargo metadata` exited **0** in that run. The stderr grep is the only signal, which
is why it is asserted separately from the lock scan.

Against the real repo it passes, in ~6 s:

```
Cargo.lock [[patch.unused]] scan:
  ok  corpan/corpan-app/src-tauri/Cargo.lock: no [[patch.unused]]
  ok  corpus-reader/src-tauri/Cargo.lock: no [[patch.unused]]
  ok  dynawalla/dynawalla-app/src-tauri/Cargo.lock: no [[patch.unused]]
  ok  panko/pako/src-tauri/Cargo.lock: no [[patch.unused]]
Declared [patch] resolution:
  ok  corpan/corpan-app/src-tauri/Cargo.toml: llama-cpp-sys-2 -> corpan/corpan-app/src-tauri/vendor/llama-cpp-sys-2/Cargo.toml

All declared patches apply.
```

### 7. Gate-code unit tests

The two new scripts are covered by the existing `ci-scripts` job
(`python -m pytest .github/scripts -q`), which this PR's own `native` filter change
also triggers. 191 tests pass, 13 of them new.

skyl:
## Review round: both MAJORs fixed, all MINORs and NITs fixed

Every finding was reproduced before it was fixed. Nothing was refuted.

### MAJOR 1 — the `[patch]` gate's non-path fail-open (`1f69e67`)

Reproduced exactly as reported. A manifest whose only patch is `{ git = ... }`:

```
patch_entries -> []
check_manifest -> []
```

Zero assertions, silent pass. End-to-end against a real git patch (local bare repo,
so no network), with the `[[patch.unused]]` stanza stripped from the lock:

```
$ python3 patch_integrity.py Cargo.toml     # BEFORE
Declared [patch] resolution:
  (no manifest declares a path-based [patch])   <- never inspected

$ python3 patch_integrity.py Cargo.toml     # AFTER
::error::Cargo.toml: cargo reported an unapplied patch: warning: patch
  `unicode-xid v0.2.99 (file:///.../vendorrepo#73a5cbdf)` was not used in the crate graph
::error::Cargo.toml: `cargo metadata --locked` exited 101 — the graph could not be
  resolved against the committed lock.
gate exit=1
```

Detection is now split from resolution, as prescribed: `has_any_patch` (any registry,
any spec shape) selects the manifest for the `cargo metadata` run; `patch_entries`
stays path-only and feeds only the vendored-directory assertion. `main` uses
`has_any_patch` for `patching` too.

**The reviewer is also right about the exit code**, and the body and docstring were
wrong. Measured on cargo 1.93.1, both directions:

| lock state | `cargo metadata --locked` | signal |
|---|---|---|
| already carries `[[patch.unused]]` | exit **0** | stderr text only |
| clean, would need rewriting | exit **101** | stderr text + `cannot update the lock file` |

So "exits 0 either way" was false, but the stderr grep is still load-bearing — it is
the *only* metadata signal in the first row. Both rows now fail, and a non-zero exit
no longer raises past the stderr warnings it would have swallowed.

`test_non_path_patches_are_skipped` pinned the hole and is gone. Replaced by
`test_git_only_patch_still_gets_a_metadata_run` and
`test_locked_failure_is_reported_not_swallowed`. Non-tautological — re-gating
`check_manifest` on `patch_entries` fails exactly those two:

```
FAILED test_patch_integrity.py::test_git_only_patch_still_gets_a_metadata_run
FAILED test_patch_integrity.py::test_locked_failure_is_reported_not_swallowed
2 failed, 198 passed
```

### MAJOR 2 — clippy had no ratchet (`76b2f9f`)

Correct: the total was written to the step summary and discarded. `CLIPPY_BUDGET=16`
is now bidirectional, in the same shape as the rustfmt `DEBT` list, at zero extra
runtime. Verified in both directions:

```
total=16 -> exit 0   at budget
total=17 -> exit 1   ::error::clippy warnings rose to 17, over the budget of 16
total=15 -> exit 1   ::error::clippy warnings dropped to 15. Lower CLIPPY_BUDGET to 15
total=0  -> exit 1   ::error::clippy warnings dropped to 0. Lower CLIPPY_BUDGET to 0
```

This run: `TOTAL unique clippy warnings: 16 (budget 16)`.

### MINOR 3 — `diagnostic_key` ignored package identity (`6582061`)

Reproduced: two records with distinct `package_id`s at an identical
`src/lib.rs:42:5` gave `(1, 0)`, and the second's rendered text never reached the
log. `package_id` is on the envelope, so the key now takes the whole record.

Confirmed against real cargo output that `--all-targets` re-emits a lib diagnostic
for the lib-test target under the **same** `package_id`, so cross-target dedup is
untouched. Both behaviours are pinned, and both mutations fail:

- dropping `package_id` from the key → `test_identical_spans_in_two_packages_count_twice` fails
- making the key unique per record → `test_same_warning_from_two_targets_counts_once` and
  `test_same_package_same_span_still_counts_once` fail

**The budget does not move.** All 16 warnings were traced to corpan-app's own
sources at 16 distinct spans (`src/lib.rs` ×6, `src/content_packs.rs` ×4,
`src/offline_cache/` ×4, `src/phrase_packs.rs` ×1, and a 16th), so no cross-package
collision was being suppressed. Dynawalla's tree is at 0. CI confirms: still 16.

### MINOR 4/5/6 + NITs — comments, which in this file are the specification

- **`--locked` comment (was ci.yml:889).** False. `git ls-files '*Cargo.lock'`
  returns **four** — corpan, corpus-reader, dynawalla, panko. Only
  `homeschool-offline` lacks one (it has no `Cargo.lock` on disk in a clean
  checkout at all). Rewritten to the real reasons: dormancy, and the git-branch
  plugin dep.
- **"All four ARE rustfmt-gated" (was ci.yml:198).** False. Measured locally:
  `corpus-reader` 17 diffs, `homeschool-offline` 126, `ios-share` 2 — all on the
  DEBT list, all `::warning::`, none able to fail the job. Only `panko` and the iap
  example are enforced. Now says walked-not-gated, and names which is which.
- **dynawalla-app job comment + its LFS comment.** Both said no cargo gate exists.
  This PR is that gate. Updated.
- **rustfmt baseline.** The comment said 1.8.0 (Rust 1.93); CI ran 1.9.0-stable /
  rustc 1.97.1. I have 1.8.0 locally and measured the full ratchet under it — the
  split is identical under both, same 14 dirty / 4 clean, same per-crate diff counts
  (31, 1, 3, 4, 3, 6, 4, 1, 3, 3, 2, 17, 126, 2). The comment now records that the
  toolchain is unpinned and that the one thing that can move the list without a code
  change is bumping a crate's `edition`, since `style_edition` defaults to it and
  these crates span 2021 and 2024.
- **`patch_entries` docstring** promised a `None`-path contract the code never
  implemented. Deleted; it now says plainly what the split is.
- **"~6 s"** for the `[patch]` step → **~1 s**. It was 1 s in CI and 0.9 s locally.
- **`mktemp` cleanup.** `trap 'rm -f "${seen}"' EXIT` added. Verified the trap
  preserves the step's verdict in both directions (`exit 1` stays 1, `exit 0` stays 0).
- **apt retry.** Added, 3 attempts with linear backoff. Verified it fails **closed** —
  `until` is exempt from errexit and the function returns 1 after the third attempt,
  which aborts the step:

  ```
  --- always fails (must abort the step) ---
  attempt 1 failed, retrying in 5s: false
  attempt 2 failed, retrying in 10s: false
  ::error::failed after 3 attempts: false
  SCRIPT EXIT=1      # the line after `retry false` never ran
  ```

### Queue safety, re-verified on the new head

```
required contexts on main : ["ci-gate","adversarial-review","hygiene"]
top-level jobs            : 15
ci-gate.needs             : 14
NOT AGGREGATED            : []          <- empty
needs entries not a job   : []
ci-gate if                : ${{ always() }}
native if                 : needs.changes.outputs.native == 'true'
```

No workflow with a top-level `paths:` filter was added or made required. `native` is
still a path-gated job, not a context.

### Unrelated flake worth recording

`terraform · fmt + validate` went red on the first run of this push with
`##[error]Error: read ECONNRESET` inside `hashicorp/setup-terraform@v3` — the binary
download. Steps 4-6 (init/fmt/validate) were `skipped`; nothing in this PR touches
terraform. It passed on re-run. Same class as the apt NIT above, but in a `uses:`
step, which a shell retry cannot wrap — noting it rather than fixing it here.

`ci-gate` correctly went red while it was, then green after. Full local suite:
**200 passed**.

PR #542, opened 2026-07-26
feat(dynawalla): fraction, decimal, choice and column-grid answer entry, plus the number line and balance scale

skyl:
## What

Give the app a structure-aware entry model and a surface for **all four V1
`AnswerSchema` kinds** — `fraction` (with mixed numbers), `decimal` (the
separator as a key on the existing `integer` schema), `choice`, and
`columnAlgorithm` — and render two more of the four V1 representations, the
**number line** and the **balance scale**.

## Why

The `AnswerSchema` union and its `RepId` representations were declared and mostly
unrendered: the app could draw `integer` entry and the counting board, and
nothing else. That is exactly the state CG-8 exists to police — a generator can
emit a perfectly valid `Exercise` the app physically cannot draw — and it is what
keeps the whole `frac` domain (32 skills, the largest in V1) out of reach.

`dw-curriculum check --strict-renderers`, which the release checklist runs,
**failed on `main`**: `answer:integer`, `answer:columnAlgorithm` and
`rep:counting-board` were all declared unimplemented while the app had been
drawing two of them since M2. It passes now, and the renderers behind it are real
— including the balance scale, whose beam was tipping the wrong way at first
review and is now measured in a browser rather than argued about in a regex (see
**Review round 2**).

Scope was cut deliberately. The **area model is not built** — it is not one of
CURRICULUM.md's four V1 representations, and its two meanings (fraction-of-a-whole
vs multiplication array) are different renderers that need the `frac` and `mul`
domains to pin down. The **gear train** stays declared and `implemented: false`:
it carries multiples, factors and LCM, none of which exists as content, so a
renderer for it could not be checked against a single real item. Both are the
brief's own instruction — a registered-but-shoddy renderer is worse than an
absent one, because the gate then lets skills go active behind it.

### Curriculum type changes (two, both load-bearing)

A renderer receives the **schema** and nothing else — `EntryModel.init`, `.keys`
and `.value` take an `AnswerSchema`, not an `Exercise` — so anything the surface
needs in order to draw an answer, or to decide what counts as the same answer,
has to be on the schema.

- `AnswerSchema.fraction.equivalence?: "as-written" | "any-equivalent"` answers
  "is `2/4` the same answer as `1/2`?" It is a per-skill curriculum decision, not
  an arithmetic one: on `simplify-to-lowest-terms`, accepting `2/4` marks the
  thing being taught as correct. Absent means `as-written`. `answerAccepted`
  implements it, and **`GeneratorFamily.check` calls it** — it did not at first
  review, which made the field a knob nothing consulted.
- `AnswerSchema.choice.options` is the drawn list itself, index-aligned with
  `AnswerValue.choice.index`. Without it a choice item is a number `k` and nothing
  to put on the screen, and the shuffle deciding which slot holds the correct
  answer has nowhere deterministic to live. `ChoiceOption` is deliberately only
  `number | fraction`: a word option would be a `LocKey`, there is no i18n runtime
  to resolve one, and a renderer that accepted `term` today would draw
  `dw.opt.line-of-symmetry` on a child's screen. `schemaDefect` refuses a set that
  offers the same number twice, however the two are written.

## Review round 2 — what the first review found, and what it cost

The reviewer was right on every blocking point. The arithmetic was sound; the
renderers were not, and neither were the checks that were supposed to see them.

**The balance scale drew the heavier pan higher.** CSS `rotate(a)` is
`x' = x·cos a − y·sin a, y' = x·sin a + y·cos a` in a space where y grows
*downward*, so it turns clockwise on screen. About `transform-origin: 50% 0` the
beam's left end sits at `x = −L, y = 0` and moves to `y' = −L·sin a`, so a
**positive** angle gives a smaller y and *lifts* the left end. `.dw-beam-left`
was `rotate(6deg)`, and `tilt === "left"` is the state `left > right`, whose text
alternative reads "The left pan is lower." Measured at 390 px on `balance-left`
(17 against 12): left pan top 6427.1, right pan top 6448.9 — the heavier pan drawn
21.8 px above the lighter one, in the one representation whose entire job is that
`=` is a relation, contradicting its own text alternative. The four rules are
swapped. Now measured: `balance-left` 3556.9 / 3535.1, `balance-right`
3768.1 / 3789.9, `balance-level` 3313 / 3313.

**The Check plate went live and inert.** `EntryModel.complete(state)` took no
schema, so it could not ask what `value(state, schema)` refuses on: it said `2/0`
was committable while `value` returned `null` at the zero-denominator guard, and
said an out-of-range choice ordinal was committable while `value` refused the
index. `committable()` read the first and `commit()` read the second, so a child
tapped Check and nothing happened at all — no verdict, no hint, no undo, nothing
to read. The method is deleted. `entryComplete(model, state, schema)` is
`value(…) !== null`, `committable` is `submitted(state) !== null`, and the two
now agree by construction rather than by having been written on the same day.

**The choice group was not keyboard operable.** `role="radio"` on a `<button>`
gets no arrow navigation from the platform — the ARIA pattern says the author
supplies it — and there was no key handler, so ArrowDown from option 1 left focus
on option 1 and a keyboard user could only ever select the first option. With a
stale out-of-range ordinal every option was `tabIndex=-1` and the group had no tab
stop at all. Arrows (both orientations), Home, End and a bounded `1..k` are
handwritten on the group now, moving relative to the **focused** option rather
than to a render closure that may not have committed — taking it from `selected`
sent ArrowDown to option 4. The roving stop is keyed off a bounded index. The
unbounded digit shortcut left the entry model, which has no schema to bound it
with.

Also fixed, from the same review:

* **`answerAccepted` was dead.** `columnOpFamily.check` compared with
  `answerEquals`, so `equivalence: "any-equivalent"` would have been silently
  ignored by the first fraction family and every child writing `2/4` marked wrong.
  `check` routes through the schema, with a test whose assertion can only pass if
  it does.
* **`schemaDefect` accepted two options of the same value** — `1/2` twice, or
  `1/2` against a decimal option worth `0.5`. Two right answers on one card.
  Rejected by exact-rational comparison, and **CG-11 now runs `schemaDefect` over
  generated items**, so the hole is measured before a generator can fall into it.
  The React key is the index, not the written option.
* **The regrouping cells were 32 × 32 px flat** (`min-h-8 min-w-8`), off the token
  scale, so they neither came down with the surface nor were covered by the cell
  test — the smallest target on the surface, and the one a six-year-old writes
  their first borrow into. On `--dw-cell-*` now, and the floor is derived instead
  of asserted (below).
* **Both grid rows labelled their cells identically**, so a child stepping cell by
  cell heard "1000, 100, 10" and then "1000, 100, 10, 1" with nothing to say which
  row. The mark row composes the existing `regroup` string; no new string.
* **`writeLinePosition(-3, 1, 4)` wrote "-3 1/4"**, which `fractionRational` reads
  as `-3 − 1/4 = -13/4`; the position is `-11/4`. Half a unit out, in the one
  notation a child would type back. It is `-2 3/4` now, and the whole function is
  checked by a round-trip property over every spec `repSpecDefect` admits in a
  sweep (3,393 positions) rather than at three non-negative points.
* **Two float divisions on the drawing path** are integer arithmetic. The guard
  meant to catch them was written against a literal string the file never
  contained; it rejects any spaced `/` outside a comment now.
* **`writeChoiceOption` had no coverage** — the test claiming it inlined a second
  copy of its body, so the function could have returned `""` with the suite green.
  It calls the function, fallback branch included.
* `assert.notEqual(cap.kind, "commit")` was true of every value `KeyCap` admits.
  Replaced with the property it was about; the commit wiring is asserted in
  `surface.test.ts`, which can see it.
* `drive:schemas` is an npm script, and the target-size checks run at 320 px,
  where the cells are smallest, and not only at 390 px.

## Blast radius

**Areas touched:** dynawalla (`curriculum/src/**`, `dynawalla-app/src/**`, docs)

Nothing under `.github/`, `dynawalla/engine/` or `corpan/` is touched. The live
practice loop is unchanged in behaviour: the ladder still serves `FORM_FREE_ENTRY`
only, so every card is still an `integer` schema on the slate. `drive-locate.mjs`
still reaches the Stage-2 contrast pair and the repair item.

- [x] Every touched path is covered by a `ci.yml` area filter
      (`dynawalla-app`, `dynawalla-curriculum`).
- [x] **Strings.** Ten new strings in `src/app/strings.ts`, nine of them text
      alternatives (`Q-10`, CG-18) — a fraction reads as "3 4" without them, a
      column grid as loose digits, a number line and a scale as nothing at all.
      The tenth, `nextField`, labels the key that moves from a numerator to a
      denominator. `surface.test.ts` pins the exact list at 24 so the count
      cannot creep. There is no locale runtime in this app yet (PR-1.6). The
      review round adds **none**: the regrouping cell's row name composes the
      `regroup` string that already existed.
- [x] **Curriculum.** No skill id renamed or reused, no node promoted, no grade
      changed. Every answer value is an exact `Rational` built through `BigInt`;
      `M-05` (no floats) passes. Two `AnswerSchema` members added — one optional,
      one required on a variant nothing constructs today. CG-11 gains a
      `schemaDefect` check over generated items (GATES.md updated).
- [x] **Native integrity.** N/A — no Rust touched.
- [x] **Pack back-compat.** N/A — no pack touched.
- [x] **Secrets.** None.

### Accessibility, per the brief

Every renderer has a text alternative, honours `prefers-reduced-motion`, and fits
320 px with no horizontal overflow. **Every renderer is keyboard operable** — the
choice group was not at first review and is now: arrows in both orientations,
Home, End, bounded `1..k`, exactly one tab stop whatever the selection is, and
`drive-schemas.mjs` presses those keys in a real browser and fails when focus does
not move.

**Target sizes, with the number derived.** The keypad's own bar is 2 cm — 75.6 px
at the 96 dpi CSS reference — and it cannot be met on the horizontal axis of a
grid cell: six columns and the decimal gutter must fit inside 320 px less the
frame's 16 px either side, so `6w + 8 ≤ 288` caps a cell at 46 px wide and the row
would then fill the frame edge to edge. So the cell is held to **WCAG 2.2 AA
2.5.8 (24 × 24 CSS px)** on both axes, and on the vertical axis — where nothing
competes for the space — to 40 px, the line box `--dw-numeral-size` needs at the
shortest rung. `surface.test.ts` checks that at every rung of the token scale, for
**both** cell tones, and `drive-schemas.mjs` measures the drawn cells at 320 px:
answer cells 32 × 47, regrouping cells 32 × 44, choice options 56 tall. The
earlier claim that the cells used "the same vertical scale the keypad uses" was
true of the scale and not of the bar, and the regrouping cells were on neither.

**One trap closed.** Every cell of a multi-field answer is a `<button>` — the only
element every platform's keyboard, screen reader and switch control already knows
how to reach — and `Practice.tsx`'s Enter handler skips buttons, because Enter on
a plate is that plate's activation. So a keyboard user who tabbed into a numerator
was *inside the answer with no way out*. `data-dw-entry` exempts entry controls,
and `surface.test.ts` asserts the exemption exists. Same class of bug as the Enter
that trapped a keyboard user on a card with no entry at all.

### What looking found, and what looking missed

`src/preview/` is a development-only renderer bench — every schema and every
representation, drawn from a real `AnswerSchema` and driven by the real entry
model. It is not a route and it is not in the shipped bundle (`vite build` inputs
`index.html` and nothing else; verified — no `preview.html` in `dist/`, no bench
code in the bundle), but it lives under `src/` so `tsc` and `eslint` cover it.
`npm run drive:schemas` drives it headlessly.

Three defects were visible in the screenshots and found there:

1. **A phantom answer rule and a second live region above the column grid.**
   `ProblemSlate` drew its integer answer row and its verdict well on every card,
   so a column card had an empty brass rule between the subtraction bar and the
   grid, and *two* verdict wells — one of which would have announced the answer to
   a card whose answer was somewhere else. Fixed by splitting out
   `ProblemStatement`; `VerdictWell` is now one component, shared.
2. **The mark row sat half a cell left of the digits it annotates** — a picture of
   the wrong procedure. Both rows now come from one `columnCells` table.
3. **The balance beam pivoted about its own corner**, swinging out over the
   heading with the fulcrum stranded below the pans and the numerals tilted. The
   beam now rotates about the fulcrum and each pan counter-rotates by the same
   angle, so the pans hang level and stay attached with no offset computed
   anywhere.

**A fourth was in the same screenshots and I did not see it**: the beam tipping
the wrong way, in `01-light-390` and `05-reduced-motion`, both of which I said I
had looked at. Looking is not a method. So the driver measures instead — it reads
`getBoundingClientRect()` off the drawn pans and fails when the heavier one is not
the lower one, reads the number line's index as a fraction of the rule and fails
when it is not standing on its own tick, presses the arrow keys at the choice
group, checks the Check plate against `2/0`, and re-runs every target-size check
at 320 px. `representation.test.ts` keeps only what a browser cannot see, plus a
*signed* angle assertion with the rotation derived in the comment rather than
copied out of the file it is checking. The mirror-image assertion it had before is
satisfied exactly as well by the inverted pair as by the correct one.

## Proof

```
$ (dynawalla-app) npm test
ℹ tests 255   ℹ pass 255   ℹ fail 0          # 195 on main; 249 at first review

$ (dynawalla-app) npm run tsc && npm run lint && npm run build
(clean)  dist/assets/index-CLQgsdP6.js 398.17 kB │ gzip: 128.29 kB
$ ls dist | grep -i preview || echo "no preview.html in dist"
no preview.html in dist

$ (curriculum) npm test
ℹ tests 97   ℹ pass 97   ℹ fail 0

$ (engine) npm test
ℹ tests 120  ℹ pass 120  ℹ fail 0            # untouched, run to prove it

$ (curriculum) npm run check
  pass    CG-8   renderer ownership
  pass    CG-11  checker self-consistency and drawable schemas
  pass    M-05   exact arithmetic only
OK

$ (curriculum) npm run check -- --strict-renderers
OK                                            # was: FAILED, 7 findings on main
```

### The real browser (`npm run drive:schemas`, 45 checks, Chrome headless)

Every check passes. The ones the review round added:

```
fraction after typing 2/0 → {"complete":"false","value":""}
OK   2/0 is not committable, so the Check plate stays dark

column cells: ["Regrouping 1000:_","Regrouping 100:_","Regrouping 10:_","1000:2","100:2","10:0","1:3"]
OK   the digit cells are named by their place, most significant first
OK   a regrouping cell says which row it is in, not only which column

OK   ArrowDown from option 1 moves focus to option 2 (got 2)
OK   ArrowRight from option 1 moves focus to option 2 (got 2)
OK   ArrowUp from option 1 moves focus to option 4 (got 4)
OK   Home from option 1 moves focus to option 1 (got 1)
OK   End from option 1 moves focus to option 4 (got 4)
choice after pressing 3 on the group → {"kind":"choice","index":2}
OK   a digit selects by ordinal, bounded by the option count
choice after a window '9': tabIndex [-1,-1,0,-1]
OK   the group has exactly one tab stop (-1, -1, 0, -1)
OK   a digit that names no option leaves the answer alone

balance-left: pan tops 3556.9 / 3535.1
OK   balance-left draws the heavier pan lower (3556.9 / 3535.1)
balance-right: pan tops 3768.1 / 3789.9
OK   balance-right draws the heavier pan lower (3768.1 / 3789.9)
balance-level: pan tops 3313 / 3313
OK   equal pans hang level
line-quarters: the index sits at 0.75 of the rule, and stands on 3/4
line-thirds:   the index sits at 0.8333 of the rule, and stands on 5/6
line-wholes:   the index sits at 0.4999 of the rule, and stands on 3/6
OK   … draws the index on its own tick

320 px column-borrow cells: Regrouping 1000=32x44, … 1000=32x47, 100=32x47, 10=32x47, 1=32x47
OK   column-borrow: every cell is a target a child can hit at 320 px
OK   column-carry: every cell is a target a child can hit at 320 px
OK   fraction / mixed: every cell is a target a child can hit at 320 px
OK   320 px: every option is still a target (56, 56, 56, 56)
```

### The loop still works (`npm run drive:locate`)

**These numbers move.** The session seed runs off the day index, which
`store.ts`'s `today()` takes from the wall clock, so the item the driver climbs to
is a different subtraction on a different day — the transcript quoted in the first
version of this body was an example read as an expected output, which was my
mistake, and the tool's header now says so. What is asserted is the shape: a run
of zeros in the minuend, a contrast board that does not add back up, a repair item
that borrows through a zero. Two runs today:

```
reached 907 − 719  correct 188  answering 288 (= correct + 100)
OK  the board shows 288 + 719 = 1007, against 907
OK  the repair item borrows through a zero: 307 − 168, the same step, correct 139

reached 703 − 579  correct 124  answering 224 (= correct + 100)
OK  the board shows 224 + 579 = 803, against 703
OK  the repair item borrows through a zero: 406 − 229, the same step, correct 177
```

### The answer path did not get slower

`tools/bench-answer-path.mjs 2000`, three runs each side, both measured on this
machine in the same session — the pre-review commit (579d1544) in a detached
worktree against this branch:

| `commit` p50 | run 1 | run 2 | run 3 |
|---|---|---|---|
| 579d1544 (first review) | 0.0086 ms | 0.0086 ms | 0.0087 ms |
| this branch | 0.0087 ms | 0.0085 ms | 0.0086 ms |

Flat. `committable` now builds the answer value instead of testing a regex, so
`commit` was calling `submitted` twice — that is one call now, with the guards
inline. (The 0.0068 ms in the first version of this body was a quieter machine,
not a faster tree: 579d1544 measures 0.0086 ms today.)

## Diff size

`git diff origin/main...HEAD | wc -c` = **248,822**, over the 200,000 the
adversarial reviewer truncates at (199,997 at first review; the round-2 fixes and
their tests added the rest). If a reviewer would rather see it split, the
natural seam is the four answer schemas (`entry.ts` + `ui/`) versus the two
representations (`NumberLine`, `BalanceScale`, `representations.ts`,
`render/representations.ts`).


skyl:
Superseded by the architecture pivot (#545), which removes `src/work/` entirely — the host now ships zero content and packs are the product.

**This work is not discarded.** The entry models and representations in this PR are exactly the primitives the pack SDK/kit needs, and should be salvaged into `dynawalla/packs/shared/` or `dynawalla/kit/` rather than rebuilt:

- `fraction` entry (numerator/denominator, mixed numbers, `2/0` refused at `value()` not at the key, `2/4` kept exactly as written)
- `decimal` entry where **the separator is a key, not a position the app fills in** — the previous model scaled by `decimalPlaces`, so a child who meant `3.5` and one who meant `35` typed the same two keys and both were marked right
- `choice` as a real radiogroup with roving tabindex
- `columnAlgorithm` — digit cells units-first with mark cells above every column but the units; a digit into a full cell passes left, the way a hand moves
- the number line and the balance scale

**Two confirmed bugs to carry forward, both found by measurement — do not reintroduce them:**

1. **The balance scale draws the heavier pan HIGHER.** `.dw-beam-left { rotate(6deg) }` raises the left end (CSS rotate is clockwise in a y-down space), but `tilt === "left"` is the state where left > right and is labelled "The left pan is lower." Measured in the rendered page: with left 17 / right 12, the 17 pan sat 21.8px **above** the 12 pan. This matters more than a normal visual bug — the balance scale is supposed to be the physical embodiment of the equals sign, so an inverted one teaches the opposite of the intended concept.
2. **`writeLinePosition` is wrong for negative origins.** For `from = −3, mark = 1, denominator = 4` the point is −11/4, but it emits "−3 1/4", which `fractionRational` reads as −13/4. `repSpecDefect` permits `from < 0` and no test covers it.

Closing rather than merging, since `src/work/` no longer exists on the target.

PR #546, opened 2026-07-26
dynawalla: eight generator families across five domains — rebased onto the stripped host

skyl:
Rebased onto the stripped host (`83a603cd0`). The eight generator families are the
point of this PR and they all survive; everything that touched the deleted
worksheet surface is gone.

## Why this exists

The founder's loudest complaint about the first build was that everything was
addition and subtraction. This is what fixes that. The curriculum library now
generates **eight families across five new domains** — place value, comparison and
ordering, rounding and estimation, multi-digit multiplication, long division,
fraction equivalence, fraction arithmetic and missing-operand equations, over
`ns`, `mul`, `div`, `frac` and `alg`.

Also: two answer schemas the judge was missing (`fraction` with the per-skill
equivalence decision, `choice` with the drawn options on the schema), thirteen new
mal-rules taking the table from 3 to 16, CG-8's missing prompt half, CG-11 running
`schemaDefect` over generated output, and `promotionBlockers.ts` naming per row
what stands between the 30 draft rows and `active`.

## What the rebase dropped

`#545` deleted `src/work/**` and `tools/drive-locate.mjs`, which is what made this
branch conflict. Rather than resurrect the worksheet, every hunk that touched it is
dropped:

- `dynawalla-app/src/work/**` — the entry models, the keypad, the four answer
  surfaces, the number line and balance renderers, and their tests.
- `dynawalla-app/src/preview/**`, `preview.html`, `tools/drive-schemas.mjs` — the
  renderer bench and its browser driver. They drove `src/work/`.
- `tools/drive-locate.mjs` — drove the deleted practice loop.
- `src/app/strings.ts`, `src/design/tokens.css`, `package.json`, `README.md` — the
  entry-surface strings, the `--dw-cell-*` tokens and the scripts that fed them.

**The host diff against `main` is empty.** `git diff origin/main..HEAD -- dynawalla/dynawalla-app`
returns nothing.

## What that forced, and what was found doing it

**The registries were claiming renderers that no longer exist.** Every entry in
`render/registry.ts` read `implemented: true` with a `testRef` into
`dynawalla-app/src/work/` — `entry.test.ts`, `representation.test.ts`,
`diagnosis.test.ts`, `surface.test.ts` — and all four are gone. CG-8 would have
stayed green pointing at nothing, which is exactly the rubber stamp
`owner`/`implemented`/`testRef` exist to prevent. Every renderer and prompt
declaration is `implemented: false` now, including the two column-op prompt keys,
because **no module in this repository draws a curriculum item today**.
`--strict-renderers` therefore fails, and that is correct rather than a regression:
a release in which nothing draws a question is not a release. The first pack
renderer clears it.

**The CG-8 tests were reading their healthy case off the shipped data** —
`assert.equal(cg8(context()).status, "pass")` was a claim about the registry being
full, so emptying it would have quietly inverted them. Both the satisfied case and
the violation are constructed now, and the shipped state is asserted explicitly:
warn by default, fail under `--strict-renderers`.

**CI was not running any of this.** The `dynawalla_curriculum` path filter matches
`^dynawalla/(curriculum|engine)/`, but the library lives at
`dynawalla/packs/shared/curriculum/` — `dynawalla/curriculum/` kept only the
devDependencies, the commands and the CLI. Every line of generator, mal-rule and
gate source sat outside the pattern that decides whether the gates run at all. A PR
that rewrote the whole library would have skipped its own job and reported green.
**This one would have.** Filter fixed.

## The two #542 bugs cannot return

Both lived in `dynawalla-app/src/work/`, which this branch no longer contains
(`grep -rn writeLinePosition` and any `BalanceScale`/`NumberLine` component return
nothing repo-wide). But deleting a bug is not the same as preventing it, and both
were *renderer* bugs that no amount of reading caught — so the two facts move into
the library every pack reads, with a test:

- `numberLinePoint(from, mark, denominator)` returns `from + mark/den` as an exact
  `Rational`. The composition that wrote `"-3 1/4"` for the point −11/4 (which
  parses back as −13/4) is not expressible. `representations.test.ts` pins every
  point on a negative line, and asserts the wrong reading is a *different number*.
- `balanceLowerPan(left, right)` states which pan hangs lower, **in weight rather
  than in an angle**, because the angle is where it went wrong: CSS `rotate()` is
  clockwise in a y-down space, so a positive angle raises the left end and the
  heavier pan was drawn 21.8 px *above* the lighter one. The test also states why
  "the two beam angles are mirror images" is not evidence — the inverted pair are
  mirror images too.

## Gates — real output

```
$ npm test
ℹ tests 156   ℹ pass 156   ℹ fail 0

$ npm run check
dw-curriculum check — incremental, 200 seeds per level
nodes=37  activeNodes=7  levels=22  generatedItems=4400  malRules=16  families=9
  pass CG-1  pass CG-2  pass CG-3  warn CG-4  pass CG-5  pass CG-6  warn CG-7
  warn CG-8  pass CG-9  pass CG-10 pass CG-11 warn CG-12 pass CG-13 pending CG-14
  warn CG-15 pass CG-16 pass CG-17 pending CG-18 pass CG-19 pending CG-20
  pending CG-21 pass CG-22 pass M-05
OK

$ npm run check:full
dw-curriculum check — full, 1000 seeds per level
nodes=37  activeNodes=7  levels=22  generatedItems=22000  malRules=16  families=9
  … same table …
OK
```

Property sweep, from `npm test`:

```
# sweep: 56500 generated cases over 113 levels (22 active, 91 draft) at 500 seeds each
# prompt registry: 25 template(s), all emitted
# 27 level(s) below CG-10's floor of 975 — none is active, and none may be promoted without reconciling it
# generator invariants: 1500 to-improper item(s) off tenths, 1500 decimal comparison(s) with the longer number smaller
# hint ladders: 7500 gen.frac.arith item(s), 3184 with a simplifying rung
# generate(): 56500 calls, p95 19 µs, p99 98 µs
```

Every `warn` is a fact about the graph, not a defect: CG-7/CG-12/CG-15 warn because
the eight new families bind `draft` rows only, and CG-8 warns because nothing draws
yet. `npm run tsc` is clean. CG-16 output hashes are unchanged.

## What still blocks a child seeing any of this

All 30 draft rows wait on a renderer, and that renderer is now a **pack's** to
land, not the host's. Eighteen of them wait on a second thing —
`promotionBlockers.ts` names them — because they do not clear CG-10's
variant-space floor, and some cannot: `dw.alg.equality.missing-addend` L0 draws
both operands from 1..9, so its true space is 81 items and no generator work
changes that. A promotion PR planned as "flip `status`" fails on the first run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,15 @@ jobs:
           # EG-1 and M-05, but there is no CG-19 equivalent in it — so a
           # bare-string prompt added under `engine/src` failed the curriculum job
           # while the engine job stayed green, on a PR that ran neither.
-          set_output dynawalla_curriculum '^(dynawalla/(curriculum|engine)/|\.github/workflows/ci\.yml$)'
+          #
+          # `dynawalla/packs/shared/curriculum/` is where the library ITSELF lives.
+          # It moved there so packs could import it without reaching into the host
+          # app's tree, and this filter did not move with it: `dynawalla/curriculum/`
+          # kept only the devDependencies, the commands and the CLI, so every line
+          # of generator, mal-rule and gate source sat outside the pattern that
+          # decides whether the gates run at all. A pull request that rewrote the
+          # whole library would have skipped its own job and reported green.
+          set_output dynawalla_curriculum '^(dynawalla/(curriculum|engine|packs/shared/curriculum)/|\.github/workflows/ci\.yml$)'
           # Not the mirror image: the engine imports nothing at all (EG-1 asserts
           # exactly that), so a curriculum change cannot break the engine job.
           # `boundary.test.ts` duplicates the float scan for this reason — "the two
--- a/dynawalla/docs/CURRICULUM.md
+++ b/dynawalla/docs/CURRICULUM.md
@@ -160,7 +160,10 @@ V1 concentrates where the evidence is genuinely deep:
 larger in every column),
 `mis.add.borrow-across-zero` (602 − 437 = 265, and 5,001 − 2,798 = 3,203 — regrouped all
 the way down, never decremented the leading digit), `mis.add.misaligned-columns`,
-`mis.add.carry-dropped`, `mis.mul.makes-bigger`, `mis.mul.partial-product-misaligned`,
+`mis.add.carry-dropped`, `mis.mul.makes-bigger`, `mis.mul.partial-product-misaligned`
+and `mis.mul.forgot-the-shift` (both children of one `mul.shift-not-applied` parent —
+the same place-value slip, but `47 × 100` answered as `47` is not a row misplaced and
+does not get the two-row repair),
 `mis.div.divisor-must-be-smaller`, `mis.div.remainder-dropped`,
 `mis.frac.add-numerators-and-denominators`,
 `mis.frac.larger-denominator-larger-fraction` (both children of one
--- a/dynawalla/docs/GATES.md
+++ b/dynawalla/docs/GATES.md
@@ -32,7 +32,7 @@ Published before M4's first curriculum PR.
 | **CG-8** | **Renderer ownership** — merge blocker | PR-4.4 |
 | **CG-9** | Level coverage by **running** generators over N seeds, ≥`minVariants` distinct | PR-4.6 |
 | **CG-10** | Variant-space adequacy — <2% duplicates over 1,000 draws | PR-4.6 |
-| **CG-11** | Self-consistency — the family's own checker accepts `canonical` and every `alsoAccept`, and rejects every distractor, over 1,000 seeds | PR-4.6 |
+| **CG-11** | Self-consistency — the family's own checker accepts `canonical` and every `alsoAccept`, and rejects every distractor, over 1,000 seeds; and every generated item's `AnswerSchema` passes `schemaDefect` | PR-4.6 |
 | **CG-12** | Mal-rule fidelity — ≥95% divergence from the correct answer | PR-8.1 |
 | **CG-13** | **Choice-laundering ban** — no `conceptual`/`reasoning` skill binds a choice-only generator | PR-4.4 |
 | **CG-14** | **Locale round-trip** — every `canonical`/`alsoAccept` survives format→parse in all launch locales; every count slot declares its CLDR plural set | PR-4.7 |
--- a/dynawalla/packs/shared/curriculum/CHANGELOG.md
+++ b/dynawalla/packs/shared/curriculum/CHANGELOG.md
@@ -9,6 +9,45 @@ installed pack until that pack is rebuilt and republished.
 
 ## 0.1.0 — Unreleased
 
+### Eight generator families across five new domains
+
+The library taught addition and subtraction and nothing else. It now generates
+place value, comparison and ordering, rounding and estimation, multi-digit
+multiplication, long division, fraction equivalence, fraction arithmetic and
+missing-operand equations — `gen.number.place-value-decompose`,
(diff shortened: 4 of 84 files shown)


skyl:
**Rebased onto `origin/main` @ `84cb373fa`, and the base retargeted from `agent/dynawalla/answer-schemas` to `main`.** #542 was closed rather than merged, so its four commits are part of this PR now. Fourteen commits replayed, plus one resolution commit.

## The relocation

`dynawalla/curriculum/src/` became `dynawalla/packs/shared/curriculum/src/` in #543. Git's directory-rename detection put every added file at the new path; the only hand resolution was `src/index.ts`, which on trunk is now two files — the compatibility shim at the old path and the real library index at the new one. Both of this branch's `index.ts` hunks were applied to the library index; the shim is untouched.

Checked rather than assumed — every source file under the old path on the pre-rebase branch, hashed against the same file under the new path on the rebased branch:

```
MISSING FROM NEW:   (none)
ADDED IN NEW:       board/countingBoard.test.ts, board/countingBoard.ts,
                    board/problem.ts, boundary.test.ts
DIFFERS:            index.ts, validate/cli.ts, validate/gates.test.ts,
                    validate/gates/generatorGates.ts, validate/gates/lintGates.ts,
                    validate/lints/scan.ts
```

The four additions and all six differences are trunk's, not this branch's: the counting board, and the relocation-aware `LINT_ROOTS` / `EMPTY_ROOT_MESSAGE` work that makes a moved lint root fail instead of scanning nothing.

## One resolution commit

`5e4a675a1` narrows the import scanner in `boundary.test.ts`, which arrived on trunk after this branch was written. It matched `\bfrom` inside string literals, so

```ts
[REP_NUMBER_LINE]: ["from", "to", "denominator", "mark"]
```

in `render/representations.ts` read as `from` followed by a quoted `, `, and the file was failed for importing the bare specifier `, `. The two trees are green apart and red together. A lookbehind now rejects a keyword whose preceding character is a quote — strictly narrower, and the three boundary tests still pass over all 103 library source files.

## Proof, re-run after the rebase (Node 24.18.0)

`dynawalla/curriculum`:

```
npm test        # tests 152  pass 152  fail 0
                #   32500 generated items across 13 configurations
                #   6500 items regenerated and compared byte for byte
                #   499/500 seeds differ between two skills at the same parameters
npm run tsc     # clean
npm run check   # ... pass CG-16, CG-17, CG-19, CG-22, M-05 ... OK
```

`dynawalla/dynawalla-app`:

```
npm run lint    # clean
npm run tsc     # clean (app + test + node projects)
npm test        # tests 316  pass 316  fail 0
npm run build   # ✓ built in 148ms
```

The two browser drivers, against a real headless Chrome:

```
node tools/drive-schemas.mjs
  balance-left:  pan tops 3556.9 / 3535.1
  OK   balance-left draws the heavier pan lower
  balance-right: pan tops 3768.1 / 3789.9
  OK   balance-right draws the heavier pan lower
  OK   equal pans hang level
  OK   line-quarters / line-thirds / line-wholes draw the index on their own tick
  OK   every cell is a target a child can hit at 320 px
  OK   the fraction bar is still drawn in dark
  OK   the beam arrives tilted rather than swinging to it

node tools/drive-locate.mjs
  OK  the board shows 337 + 365 = 702, against 602
  OK  the repair item borrows through a zero: 700 − 671, the same step, correct 29
```

Both bugs #542 was closed over stay fixed, and both are measured rather than read:

- the balance draws the **heavier** pan lower, in both directions, at real pixel coordinates;
- `writeLinePosition(-3, 1, 4)` is `-2 3/4` — asserted directly and swept against `fractionRational` over the whole position space in `representation.test.ts`.

PR #549, opened 2026-07-26
FORGE — an incremental smelting chain where the arithmetic is the throttle

skyl:
**A furnace you feed with arithmetic.** Six stations, each one building the station
below it. Bellows make sparks, crucibles make bellows, hammers make crucibles, all
the way up to the reactor — so buying deep is a fifth-order effect on the counter,
and ninety seconds later the curve you are watching is visibly steeper than the one
you were watching before.

New, self-contained package: `dynawalla/games/forge/`. Nothing outside that
directory is touched, and it imports nothing from `curriculum/` or `engine/`.

```
cd dynawalla/games/forge && npm install && npm run dev   # 127.0.0.1:1431
npm test    # 49 tests, node's runner, no framework
npm run tsc && npm run build
open http://127.0.0.1:1431/bench.html                    # frame budget
```

---

## Five maths surfaces, all native to the genre

| | |
| --- | --- |
| **THE STRIKE** | The answer **is** the payout. `12 × 11` pays 132, `4 + 5` pays 9, plus a second of your entire production, times the combo. Within a minute a child starts preferring the big ones — the first time most of them voluntarily compare two arithmetic expressions. |
| **HEAT** | `1 + √heat/10`. 100 heat is ×2.00, 2500 is ×6.00, 10 000 is ×11.00. Bleeds at 1/16 per second, so the multiplier is visibly draining while you decide. A wrong strike costs a quarter of it — **the better you are doing, the more a guess costs.** |
| **THE SEAL** | Stations 3–6 arrive chained shut; answer to crack one. Maths exactly where a free-to-play game would show a video. Wrong just re-heats the seal and asks again — unlocking is never a wall. |
| **THE FORGE MARK** | `+14 HAMMER` or `×2 HAMMER`, with the HAMMER row three centimetres away reading **9**. `C + N > 2C ⟺ N > C` — pre-algebra, and **the crossover moves as you play**: at C=3 the `+14` is obvious, at C=400 it is laughable. You keep whichever you take; the penalty for wrong is that your numbers grow slower. No red X anywhere. |
| **THE QUENCH** | Prestige pays `⌊√(lifetime / 10¹²)⌋`, and the reward screen is the radical being evaluated, in figures, with no words: `√( 4.10×10¹⁴ / 10¹² )` → `√ 410` → `= 20 CARBON`. |

Plus the offline haul, claimed with one strike — full value for a right answer,
half for a wrong one.

**Milestones fire on the number you are watching** (the highest the counter has
ever read), not on lifetime production. A `10⁹` stamp over a counter reading
`5.2×10⁷` is a lie about which number just did something — and tying the two
together produces the only real tension in the early game: spend now, or hold ten
more seconds and watch the exponent tick over.

## Exact arithmetic

Every quantity a player can see, spend, compare or brag about is a **BigInt** in
micro-units. No float in a cost, a threshold, a comparison or a displayed digit.
Growth is unbounded; the readout flips to `4.271 ×10⁹` at a million with the
exponent set in large raised figures, because from there it is the score.

`src/scene/header.test.ts` exists because the printed heat multiplier and the
applied one drifted apart once already during this branch, when heat went from
linear to square-root and only the economy was updated. **A maths game that
displays a multiplier it is not applying is the one bug this cannot ship.**

## Juice, by name

Combo-scaled hitstop (58–140 ms), decayed screenshake with a separate directional
camera punch, area-conserving squash on the billet, `outElastic` on a struck
ingot, `outBack` on a station slamming into its bay, slow-motion (0.28–0.32×) on
only the three rarest beats. Pooled `Float32Array` particles blitted against
pre-tinted radial sprites — never `shadowBlur`. Procedural WebAudio in
transient/body/tail layers with an *inharmonic* 1 : 2.76 : 5.40 : 8.93 steel body,
a pentatonic combo walk, per-hit detune, and a forge bed that audibly roars louder
as the heat climbs.

**Children's-product constraints are enforced in code, not by convention.**
`prefers-reduced-motion` removes motion and drops nothing informational; hitstop
survives, because it is timing rather than movement. Full-screen flashes are
rate-limited to 3/second and amplitude-capped (0.55, or 0.15 reduced) *by
construction* — a test fires a flash request every frame for a second and asserts
at most three land, keeping it below the 3 Hz photosensitivity band. No streaks,
no loss, no timers, no variable-ratio rewards. Stopping costs nothing.

## Measured, not asserted

`tools/bench.ts` calls the real `drawScene` synchronously in a tight loop, because
an fps counter can be throttled behind your back and a millisecond cannot.
1280×900 at dpr 2 (4.6 MP — larger than any phone this ships to):

| load | median | p95 | share of a 60 fps frame |
| --- | --- | --- | --- |
| at rest | 0.60 ms | 1.20 ms | 4 % |
| 720 particles | 0.90 ms | 5.30 ms | 5 % |
| 1100 (pool full) | 1.00 ms | 5.20 ms | 6 % |

Economy tick: **1.5 µs**, and **4.9 µs** even at 10⁴⁰⁰. Safety net is an adaptive
particle budget that sheds particles rather than frames below 52 fps; dpr is
capped at 2.

`tools/balance.ts` simulates a full session against the real economy: four
stations by 3:36, first quench at 4:54, 13 orders of magnitude in twenty minutes
at a mean gap of 58 s, 12 forge marks. The free tier is 5–10 minutes a day; in
five minutes a player meets four stations, crosses six or seven powers of ten, and
watches the quench plate light up cyan just out of reach.

## Bugs found and fixed while playtesting the real UI

- **Production only caught up on a cold start.** Switching tabs for twenty minutes
  earned nothing, because `requestAnimationFrame` stops and the catch-up only ran
  at mount. `visibilitychange` now runs the identical 1 Hz catch-up path.
- **`glow()` reset `globalAlpha` to 1** instead of restoring it, so every glowing
  label ignored the overlay fade it was drawn inside — the quench panel faded in
  from nothing while its text sat at full strength.
- **The quench plate was clickable but never drawn**, and `carbonFor` (an `isqrt`
  of a several-hundred-digit integer) was on the per-frame path. Now cached and
  recomputed twice a second.
- **`√410 = 6` was on screen** — 6 is the *gain*, 20 is the root. The line under a
  radical now says what the radical evaluates to; the gain moved to the button.
- Milestone punch and forge mark opened on the same frame, so the flash washed out
  the panel. The mark is now queued 0.95 s behind the punch.

## Verification

49 tests · `tsc` clean on source and tests · production build 62.0 kB (22.6 kB
gzipped, dev hook stripped) · played for real in Chrome at 1280×900 and 390×844,
with screenshots of the resting screen, a strike, a milestone, a failure, peak
escalation, and all four overlays.

## Reviewer note on size

`git diff origin/main | wc -c` is **~256 kB**, over `MAX_DIFF_BYTES` (200 000), so
`adversarial-review` will truncate. This is one self-contained game in one
directory and splitting it across PRs would produce halves that neither build nor
play, so I did not. To spend the review budget well, read in this order:

1. `src/core/economy.ts` and `src/core/bigmath.ts` — all the exactness claims
2. `src/game/marks.ts` — the pre-algebra comparison and its tests
3. `src/game/forge.ts` — the loop, input and every state transition
4. `src/render/juice.ts` — the flash/reduced-motion guarantees
5. `src/scene/*` — rendering only; no game rules live there

`package-lock.json` (29 kB of the diff) is kept deliberately so the next agent
gets a reproducible install.


skyl:
**Rebased onto `origin/main` @ `84cb373fa`** (was based at `2c8f68086`, four commits behind). Clean — `git range-diff` reports both commits unchanged; nothing in this PR shares a file with anything that landed in between.

Re-verified after the rebase, from `dynawalla/games/forge`:

```
npm test    # tests 49  pass 49  fail 0
npm run tsc # tsc --noEmit && tsc --noEmit -p tsconfig.test.json — clean
npm run build
            # dist/assets/index-CS2nCGB4.js  62.00 kB │ gzip: 22.56 kB
            # ✓ built in 33ms
```

skyl:
**Follow-up: `hygiene` is now green.** It had failed on this branch since it opened — on `0ef40b44a` before the rebase and on `9b71980f4` after — so it was not caused by the rebase, but it was blocking the merge.

gitleaks 8.18.4's `generic-api-key` rule was matching the save slot:

```
generic-api-key  dynawalla/games/forge/src/game/save.ts:8  KEY = "dynawalla.forge.save.v1"
```

`dynawalla.fuse.best` in `games/merge` is the same construct and passes; the difference is length, not content — the longer string clears the rule's Shannon-entropy floor, and the rule keys off an identifier called `KEY` beside a quoted literal, which is what every `localStorage` slot looks like.

Renamed to `SAVE_SLOT` and annotated `gitleaks:allow` with the reason, squashed into the FORGE commit because the scanner reads the commit range and would keep flagging the original patch otherwise. **The value is unchanged**, so saves written before this still load.

```
gitleaks detect --log-opts="84cb373fa..HEAD"   # no leaks found
npm test                                       # tests 49  pass 49  fail 0
npm run tsc                                    # clean
npm run build                                  # ✓ built in 32ms
```

CI, hygiene and adversarial-review are all green on `6cdb4eba5`.

PR #551, opened 2026-07-26
foundations(physics): bake off four engines, choose Rapier, ship the kit

skyl:
## What this is

The **physics foundation** for the Dynawalla Bazaar: a four-engine bake-off, the
decision it justifies ([ADR-0020](../blob/agent/dynawalla/found-physics/dynawalla/docs/DECISIONS/ADR-0020-physics-engine-rapier.md)),
a reference kit, and a runnable demo.

Everything in [`dynawalla/foundations/physics/README.md`](../blob/agent/dynawalla/found-physics/dynawalla/foundations/physics/README.md)
was measured on this machine on 2026-07-26 by a script in `bench/`, and every script is
committed. Where something was **not** measured — a real Galaxy Tab A9, an x86 device —
it says so rather than rounding up.

## Acceptance criteria

Moves no item to MET. It is the groundwork for **`Q-02`** (holds 60 fps on the Galaxy Tab
A9), which is `[device]`-gated and cannot be met without the device; this PR commits the
measurement rig and a budget derived from a CPU-throttled proxy, and states plainly that
the proxy is not the device. Worth landing because 300 planned prototypes will each
either inherit this or re-derive it, and three of the traps below silently produce a
**dead mechanism that looks like working code**.

## The decision

**Rapier 2D, pinned at `@dimforge/rapier2d-compat` 0.19.3.** It wins on capability before
it wins on speed — each alternative fails something on the required list:

| engine | the disqualifier |
|---|---|
| Matter.js 0.20.0 | **0 of 300 dominoes fall** at friction ≥ 0.45; wave dies at domino three. Its iteration dial makes chain stretch *worse* (17% → 62%). Last published 2024-06-23. |
| Planck.js 1.5.0 | chain stretches **64% at a 10:1 load ratio**; 8.8x slower than Rapier on 500 bodies |
| Box2D v3 (`box2d3-wasm` 5.2.0) | fastest solver (6x Rapier's median) but Embind allocates per getter — 500 transforms cost **0.454 ms vs 0.171 ms** — and it settles a balance scale at **46-70° with equal pans** |

Rapier's cost is size: 1.12 MB WASM, 2.75x Box2D v3's. Accepted, with the reasoning and
the ~200 KB gzip `-compat` → separate-`.wasm` switch documented.

## Determinism

Verified, not assumed. One seeded scene (toppling stack + loaded chain + 120 particles +
a projectile), 900 steps → **identical state hash in Node/V8, Chrome/V8 and
WebKit/JavaScriptCore**. Structural: the module imports **no libm from JavaScript**, so
every float op is inside WASM where f32/f64 semantics are fully specified. Not verified:
cross-architecture (arm64 only).

Replays therefore store **commands addressed by step index**, never positions.

## Traps, each now gated by a test

1. **Rapier joints collide the bodies they connect and expose no `collideConnected`.** A
   pendulum whose anchor has a collider does not move — **0.0° in 5 s**, vs -33.0°
   without. Every articulated recipe here would have been silently dead.
2. **`JointData.limitsEnabled` / `.limits` are accepted, type-check, and are ignored.**
   Only `joint.setLimits()` works; the wrong spelling lets a ±22° stop reach **-174°**.
   This test then caught a *second* bug — my own bound `w.pin` was dropping the argument.
3. **A beam pivoted through its own centre of mass is neutrally stable**, so the same
   scene settles at +0.05° / +1.5° / -70.0° / +72.4° across the four engines. Not four
   bugs — one under-determined scene.

Plus: `box2d3-wasm` silently serves the scalar build unless `crossOriginIsolated`;
`advance()` must clamp or a backgrounded WebView freezes for seconds on resume; Node's
`--experimental-strip-types` is strip-**only** (no parameter properties).

## The equals sign

Given its own section because the brief singled it out. Pivot raised above the beam's
centre of mass (which is also the legibility dial — measured at five heights), real
mechanical stops, pans with lips. Settles at **0.01° with 0.000° jitter** at equality and
tips to its stop under one extra unit in four.

And the rule that is not about physics: **`compare()` is an integer comparison of what
was put in the pans, never the beam angle** — correct before a single step has run, with
a test that asserts it after forcibly rotating the beam. The solver drives the picture;
it never decides a mathematical fact. This extends ADR-0009's "true by construction".

## Budget

**4 ms of a 16.67 ms frame, as step p99**, four tiers whose floor is the reference
tablet. At `mid` the heaviest scene holds 60+ fps through **12x** CPU throttling in real
Chrome; every other scene holds 120 fps at 8x.

## Review notes

- **Diff is ~242 KB.** `adversarial_review.py` on `main` **chunks** rather than truncates
  (the `dynawalla/AGENTS.md` warning describes the pre-PR-0a.2 behaviour), so this is
  fully reviewed across 2 chunks. 36 KB of it is `package-lock.json`. Kept as one PR
  because the evidence and the thing the evidence justifies are not separately reviewable.
- **No new required status check.** `dynawalla/foundations/` matches no `ci.yml` area
  filter, so it lands in the `uncovered` warning bucket by design — per
  `dynawalla/AGENTS.md`, never a fourth required context.
- **ADR number may need rebasing.** 0020 was the next free slot when this was written;
  other agents are in flight.
- **Relationship to ADR-0004.** ADR-0020 does not claim ADR-0004's revisit condition is
  met, and does not touch the separate question of whether V1 renders in 3D.

## Verify locally

```bash
cd dynawalla/foundations/physics && npm install
npm test          # 17 gates
npm run tsc
npm run demo      # localhost:1425 — six scenes, live perf HUD
node bench/run-node.mjs --repeat 3
node bench/determinism.mjs      # needs the demo running
node bench/run-browser.mjs      # real Chrome, CPU-throttled
```

## Still owed

A real Galaxy Tab A9 run; cross-architecture determinism; a check inside an actual Tauri
WebView (whether the custom protocol serves `.wasm` with a MIME type that allows
`instantiateStreaming`); and the `-compat` → separate-`.wasm` switch before shipping.

### DIFF
--- a/dynawalla/docs/DECISIONS/ADR-0020-physics-engine-rapier.md
+++ b/dynawalla/docs/DECISIONS/ADR-0020-physics-engine-rapier.md
@@ -1,0 +1,101 @@
+# ADR-0020 — Rapier 2D is the physics engine
+
+**Status:** Proposed
+**Supersedes:** nothing. **Interacts with:** [ADR-0004](ADR-0004-no-mic-no-llm-no-3d.md).
+
+## Context
+
+The Bazaar's games need real physics: projectile arcs with usable aim-assist, stacking
+and toppling, ropes and chains, soft bodies, a particle-liquid approximation, gear
+trains, dominoes, and a balance scale that is the physical embodiment of the equals sign.
+
+Four candidates were benchmarked against identical scenes rather than compared from
+documentation: **Rapier 2D** 0.19.3, **Box2D v3** (`box2d3-wasm` 5.2.0), **Planck.js**
+1.5.0 and **Matter.js** 0.20.0. The full evidence, every table, and instructions to
+re-run everything are in [`dynawalla/foundations/physics/README.md`](../../foundations/physics/README.md).
+Havok and Jolt were excluded on the record: both are 3D engines whose shipped JS surface
+is a Babylon plugin API, and the games are 2D-simulated.
+
+## Decision
+
+**Use Rapier 2D, pinned to `@dimforge/rapier2d-compat` 0.19.3**, behind the kit in
+`dynawalla/foundations/physics/src/`. Prototypes call `createWorld()` and the recipes;
+they do not talk to Rapier directly.
+
+**The physics never decides a mathematical fact.** Quantities a prototype may treat as
+truth — `compare()`, `momentOf()`, `volumeIn()`, `fallenFraction()` — are computed by
+exact arithmetic over what the game put into the world. The solver drives the *picture*.
+This extends [ADR-0009](ADR-0009-stakes-without-loss.md)'s "true by construction" to the
+simulation layer, and it is enforced by a test: `compare()` is already correct before a
+single step has run, and stays correct after the beam is forcibly rotated.
+
+**Budget: 4 ms of a 16.67 ms frame, measured as step p99**, with four quality tiers whose
+floor (`mid`) is the reference tablet, not a downgrade from desktop.
+
+## Why Rapier and not the others
+
+Each of the three alternatives failed a capability that is on the required list, not a
+performance target.
+
+- **Matter.js drops 0 of 300 dominoes** at friction 0.45 and 0.7 — the wave dies at
+  domino three. Rapier, Box2D v3 and Planck agree within 2 dominoes across the whole
+  friction sweep; Matter falls off a cliff above 0.3. Its iteration count also makes
+  chain stretch *worse* (17% → 62%), so it is not a convergence knob. Last published
+  2024-06-23.
+- **Planck.js stretches a chain 64% at a 10:1 load ratio** and 92% at 150:1, and more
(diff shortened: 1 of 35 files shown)


skyl:
**Rebased onto `origin/main` @ `84cb373fa`** (was based at `2c8f68086`, four commits behind). Clean — both commits unchanged, no shared files with what landed in between.

Re-verified after the rebase, from `dynawalla/foundations/physics` on Node 24.18.0:

```
npm test    # tests 17  pass 17  fail 0
npm run tsc # clean
```

Includes the three tests that matter for the equals sign, all green:

```
✔ THE EQUALS SIGN: equal pans settle level, and stay level
✔ THE EQUALS SIGN: an imbalance tips the right way, visibly
✔ THE EQUALS SIGN: compare() is arithmetic, not the solver
```

The second asserts a **negative** beam angle when the right-hand pan is heavier — i.e. the heavier pan goes down, measured rather than declared. The inverted balance from #542 cannot recur here: the beam is a simulated rigid body under gravity, not a CSS `rotate()`.

PR #628, opened 2026-07-29
feat(dynawalla/games/claim): the ground stops at the safe edge, and the score comes out from under the host's button

skyl:
## What

Lay CLAIM's arena and HUD out inside the safe rectangle, move the level counter,
the score and the lives out from under the host's two 44px corners, give the
mute button a real touch target off the home indicator, and add a how-to-play
surface the game has never had.

## Why

CLAIM declares `viewport-fit=cover`. Its HUD is DOM, so it *could* claw the
notch back with `env(safe-area-inset-top)` — and did. Only the top, and only for
the HUD. Four things were left out:

- **The arena.** The grid is centred in a canvas that runs to the bottom and
  both sides of the frame. Cells are not decoration: they are the ground the
  child drives across and the thing the fraction is measured in. A row of cells
  under the home indicator is ground that cannot be seen, cannot be cut, and
  swallows an upward drag as a system gesture.
- **The sides.** `padding: ... 12px` ignored the left and right insets, so in
  landscape `LVL 4` sat behind the notch.
- **The host's two corners.** The level counter was under the exit control; the
  score and the three lives were under how-to-play. A child who cannot watch
  three pink squares go to two does not know the run is ending.
- **The mute button.** `right: 10px; bottom: 10px` — inside the home indicator —
  and 34px square against a 44px platform floor.

**Why a new module rather than more CSS.** `env()` cannot be asserted from a
test. A CSS-only fix would be exactly as invisible to this package's suite as
the four bugs were, and "check it on a device" is precisely how all four
shipped. So the numbers moved into `src/game/layout.ts` — one pure function per
surface, `insets` **required**, not optional — `Hud.layout` publishes them as
custom properties, the stylesheet consumes them (keeping `env()` fallbacks for a
HUD that somehow never gets laid out), and `test/layout.test.ts` asserts the
same values through `hitsHostChrome`.

Host chrome **overlays**; nothing reserves a band. The two clusters step
**inboard** of the corners rather than downward, which costs the arena nothing —
reserving a top strip cost 12% of a small phone's height in SKY LEDGER and broke
its own layout. The fraction bar spans the full width and so cannot dodge
sideways, so the top row gets exactly enough `min-height` to put the bar below
the corners and not a pixel more; the stacked goal fraction is already taller
than that on most screens, so in practice nothing moves.

## Blast radius

**Areas touched:** dynawalla (one game pack, `dynawalla/games/claim` only)

- [x] Every touched path is covered by a `ci.yml` area filter.
- [x] **Pack back-compat.** No published zip changed in place; `pack.json`
      untouched (id, version, capabilities, the three declared `dw.frac.*` rows
      all unchanged). No `voiceId`, no catalog entry.
- [x] **Native integrity.** N/A — no Rust, no `src-tauri`, no manifest touched.
- [x] **Strings.** N/A for `corpan-app` locales. The new strings are instruction
      copy inside the pack; it declares `"locales": ["en"]` and has no i18n
      catalogue, same as SKY LEDGER's adoption.
- [x] **Curriculum.** N/A — no skill id, grade, generator or answer schema
      touched. `exact.ts`, `levels.ts` and the exact-cell band are untouched and
      `claim.test.ts` passes unchanged.
- [x] **Secrets.** None. `gitleaks` clean over `origin/main..HEAD`.

**What a reviewer cannot reconstruct from the diff:**

*The arena still owns every pixel it used to on a device with no insets.*
`arenaRect(w, h, {0,0,0,0})` is asserted to equal `{0,0,w,h}` exactly — the
safe-area work has to be a no-op where there is nothing to avoid, or it is a
regression dressed as a fix. Only the **grid** is constrained; the canvas still
covers the whole stage and the paper and vignette still bleed to the edges,
which is the entire point of `cover`.

*`stageAspect()` now measures the safe part of the stage.* `pickArena` chooses
which 7200-cell arena shape to use; every shape is exactly 7200 cells in exactly
40 blocks, so this changes *which* 7200 on a notched landscape phone, never how
many. The goal's meaning is unaffected — that invariant is what `claim.test.ts`
already guards.

*Insets are re-read on every layout, not cached.* `onInsetsChange` is subscribed
in the constructor and unsubscribed in `destroy`, because rotation swaps
top/bottom with left/right and Split View changes them outright, and neither
necessarily resizes the stage. Without it the HUD is laid out for the shape the
tablet was in when the pack opened.

*`destroy()` now calls `stopInsets()` and `guide.destroy()`*, removing the help
button, the panel, the injected stylesheet, the `keydown` handler and the inset
watcher.

## Proof

`npm test` five times — one green run is not proof:

```
$ for i in 1 2 3 4 5; do npm test 2>&1 | grep -E "^# (tests|pass|fail)"; done
--- run 1 ---
# tests 57
# pass 57
# fail 0
--- run 2 ---
# tests 57
# pass 57
# fail 0
--- run 3 ---
# tests 57
# pass 57
# fail 0
--- run 4 ---
# tests 57
# pass 57
# fail 0
--- run 5 ---
# tests 57
# pass 57
# fail 0
```

**Both halves fail when the fix is removed.** A clearance test that passes
either way is worthless.

Corner clearance — drop the gutters and the top row's min-height:

```
not ok 27 - nothing a child must read sits under the host's corners — phone portrait, small (320×568)
not ok 30 - nothing a child must read sits under the host's corners — phone portrait, tall (390×844)
not ok 33 - nothing a child must read sits under the host's corners — tablet portrait (768×1024)
not ok 36 - nothing a child must read sits under the host's corners — tablet landscape (1024×768)
not ok 39 - nothing a child must read sits under the host's corners — phone landscape (844×390)
not ok 42 - nothing a child must read sits under the host's corners — laptop (1440×900)
not ok 45 - nothing a child must read sits under the host's corners — phone portrait, notch + home indicator (390×844)
not ok 48 - nothing a child must read sits under the host's corners — phone landscape, notch on the left (844×390)
not ok 51 - nothing a child must read sits under the host's corners — small phone (320×568)
# tests 57
# pass 48
# fail 9
```

Safe area — `arenaRect` back to the full canvas, HUD padding back to a bare
12px, mute back to 34px at `bottom: 10px`:

```
not ok 29 - the mute button is a real, reachable target — phone portrait, small (320×568)
not ok 32 - the mute button is a real, reachable target — phone portrait, tall (390×844)
not ok 35 - the mute button is a real, reachable target — tablet portrait (768×1024)
not ok 38 - the mute button is a real, reachable target — tablet landscape (1024×768)
not ok 41 - the mute button is a real, reachable target — phone landscape (844×390)
not ok 44 - the mute button is a real, reachable target — laptop (1440×900)
not ok 47 - the mute button is a real, reachable target — phone portrait, notch + home indicator (390×844)
not ok 49 - the HUD stays inside the safe area — phone landscape, notch on the left (844×390)
not ok 50 - the mute button is a real, reachable target — phone landscape, notch on the left (844×390)
not ok 53 - the mute button is a real, reachable target — small phone (320×568)
not ok 54 - the arena keeps its ground out of the unsafe edges — phone portrait, notch + home indicator
not ok 55 - the arena keeps its ground out of the unsafe edges — phone landscape, notch on the left
not ok 56 - the arena keeps its ground out of the unsafe edges — small phone
# tests 57
# pass 44
# fail 13
```

Node reports zero insets, so the safe-area half runs against three *simulated*
notched devices — without those it could not fail at all.

```
$ npm run tsc
> @dynawalla/game-claim@0.1.0 tsc
> tsc --noEmit

$ npm run build
✓ 23 modules transformed.
dist/index.html                  0.56 kB │ gzip:  0.33 kB
dist/assets/index-CcKH7mjn.css   7.22 kB │ gzip:  2.30 kB
dist/assets/index-CjCE8Sh_.js   65.26 kB │ gzip: 22.67 kB
✓ built in 153ms
```

```
$ cd dynawalla/packs && node build.mjs claim
dynawalla.claim@0.1.0 — CLAIM
  entry        pack.html
  host         0.1.0 … <1.0.0
  sdk          1.0.0 (this SDK is 1.0.0)
  capabilities items, items.reveal, haptics
  covers       3 skills, grades 3–5
  on disk      4 files, 77404 bytes

1 pack(s) built, checked and staged into src-tauri/packs/
```

```
$ gitleaks detect --no-banner --redact --log-opts="origin/main..HEAD"
INF 1 commits scanned.
INF scanned ~19832 bytes (19.83 KB) in 34.6ms
INF no leaks found

$ git diff --name-only origin/main...HEAD
dynawalla/games/claim/src/game/hud.ts
dynawalla/games/claim/src/game/index.ts
dynawalla/games/claim/src/game/layout.ts
dynawalla/games/claim/src/game/render.ts
dynawalla/games/claim/src/style.css
dynawalla/games/claim/test/layout.test.ts

$ git diff --name-only --diff-filter=D origin/main...HEAD
(empty)
```

### DIFF
--- a/dynawalla/games/claim/src/game/hud.ts
+++ b/dynawalla/games/claim/src/game/hud.ts
@@ -6,7 +6,9 @@
 // the whole pedagogy, and it needs real type at real weights, which is DOM.
 
 import { cleanFraction, percentTenths } from "./exact.ts"
+import { hudFrame } from "./layout.ts"
 import { css, type LevelInk } from "./palette.ts"
+import type { Insets } from "../../../../packs/shared/game-chrome/index.ts"
 
 export class Hud {
   root: HTMLDivElement
@@ -72,6 +74,25 @@ export class Hud {
     this.card = document.createElement("div")
     this.card.className = "cl-card"
     parent.appendChild(this.card)
+  }
+
+  /**
+   * Publish the frame as custom properties the stylesheet consumes.
+   *
+   * The stylesheet keeps `env()` fallbacks on every one of these, so a HUD that
+   * never gets laid out is still inside the notch — but the values under test
+   * are these, and these win.
+   */
+  layout(w: number, h: number, insets: Insets): void {
+    const f = hudFrame(w, h, insets)
+    const s = this.root.style
+    s.setProperty("--cl-pt", `${f.padTop}px`)
+    s.setProperty("--cl-pl", `${f.padLeft}px`)
+    s.setProperty("--cl-pr", `${f.padRight}px`)
+    s.setProperty("--cl-gl", `${f.gutterLeft}px`)
+    s.setProperty("--cl-gr", `${f.gutterRight}px`)
+    s.setProperty("--cl-toph", `${f.topMinH}px`)
+    s.setProperty("--cl-cluster", `${f.clusterW}px`)
   }
 
   setLevel(level: number, ink: LevelInk, goal: { n: number; d: number; target: number }): void {
--- a/dynawalla/games/claim/src/game/index.ts
+++ b/dynawalla/games/claim/src/game/index.ts
@@ -7,6 +7,12 @@
 // a cut that pushes you past the band does not show you a red X — it falls
 // apart and gives the ground back to the hunters.
 
+import {
+  createInstructions,
+  onInsetsChange,
+  safeInsets,
+  type Instructions,
+} from "../../../../packs/shared/game-chrome/index.ts"
 import type { Host, Question } from "../contract.ts"
 import { Audio } from "./audio.ts"
 import { cleanFraction, clamp } from "./exact.ts"
@@ -35,6 +41,7 @@ import {
 import { Hud } from "./hud.ts"
 import { Input } from "./input.ts"
 import { Juice } from "./juice.ts"
+import { arenaRect, muteRect } from "./layout.ts"
 import { goalFromQuestion, levelAt, type Goal, type Level } from "./levels.ts"
 import { css, INK, levelInk } from "./palette.ts"
 import { Particles } from "./particles.ts"
@@ -148,6 +155,8 @@ class Claim {
   private fpsAcc = 0
   private fps = 60
   private worstFrame = 0
+  private guide: Instructions
+  private stopInsets: () => void
 
   constructor(el: HTMLElement, host: Host) {
     this.host = host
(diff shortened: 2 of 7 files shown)


skyl:
Force-pushed one fix found in self-review, plus a second-source-of-truth cleanup.

**The how-to-play button would have been dead to a finger.** `Input` binds `pointerdown` to the pack ROOT — unlike BEAM and COUNTERPOISE, which bind to their canvas — and it calls `e.preventDefault()` unconditionally. The help button and its panel are children of that same root, so a tap on `?` bubbled to the handler, `preventDefault()` suppressed the compatibility mouse events a `click` is synthesised from, and the panel never opened. It also anchored the steering stick and set `anyPress`, so tapping the manual dismissed a card and started driving.

`Input` now takes the game surface (`.cl-shake` — the HUD and the arena) alongside the root, and ignores any pointer that lands outside it. That covers the mute button too, which had been carrying its own `stopPropagation()` for exactly this reason.

It also gains an `enabled` flag: while the panel is open the frame loop skips `update`, so the hunt cannot close on a child who is reading the rules, the revive countdown stops, and an arrow key behind the scrim is not a move.

**The mute button is now positioned from `muteRect`** via custom properties rather than from a second copy of the numbers in the stylesheet — same shape as the HUD frame. It was the one surface in this PR the test asserted without driving.

`npm test` × 5, `tsc`, `build` and `build.mjs claim` all re-run green after the amend.

PR #634, opened 2026-07-29
dynawalla/games/siege: env() on two edges is a half-fix, and the status bar was the host's row

skyl:
## What

Adopt the shared game chrome in **SIEGE** (`dynawalla/games/siege`): honour the
safe area on **all four edges** rather than two, keep the host's two 44px
corners clear of the status bar, add a how-to-play manual, and fold in the iOS
gesture guards SIEGE was the only pack of 27 missing.

## Why

**`env()` on two edges is a half-fix.** SIEGE was one of the seven games that
already read `env(safe-area-inset-*)` — and it read exactly two of them.
`.sg-top` had `padding-top: max(8px, env(safe-area-inset-top))`, `.sg-anvil` had
`padding-bottom: …bottom`, and **neither side was touched at all**.
`viewport-fit=cover` opts a document into the rounded corners on *every* edge,
not just the one that is obvious in portrait. Held sideways the cutout is a
**side** inset — and the side of that bar is where the ember count and the sound
switch lived. A currency a child cannot read and a control they cannot press.

Now honoured on all four edges: the status bar, the console, the overcharge
overlay, the defeat screen, the banners, and the wide layout's right-hand
column. The square board is fitted inside the safe box too — `computeView` takes
the safe rect as a **required** argument, because an optional safe area is a
game that forgets to pass one, compiles clean, and is found on hardware.

**The status bar was the host's row.** The host paints an exit control over the
top-left 44px corner and a how-to-play control over the top-right one, on top of
this pack, and SIEGE's status bar runs the full width of exactly that row. The
ember count was under the first and the sound switch under the second. The bar's
*contents* now start after the first control and stop before the second, driven
by `--sg-corner` written in from the shared `HOST_*` constants. **Nothing is
reserved**: the bar keeps its height, the board keeps its size, the forge still
reaches the glass. Reserving the whole top strip is what broke a sibling game's
lattice at 320×568.

Paying 108px for those two corners meant the three switches — call wave, speed,
sound — had to leave the bar. They are in the anvil's head row now. **That bar
was over-full on a 320px phone anyway**: `overflow: hidden` was quietly cutting
those same three switches off the right-hand edge, so nobody on a small phone
could reach them at all. Moving them is not a regression; it is the first time
they have been reachable there. `TOP_BAR_MIN` is asserted so the clearance can
never squeeze the four remaining figures out.

**Gesture guards (the one extra job).** SIEGE was the only pack of 27 with no
`maximum-scale=1` and no `touch-action: none` in its entry. A double tap inside
the pack could scale and then pan the whole host document, with no way for a
child to undo it. `pack.html` and `index.html` now match `games/slice/pack.html`
exactly bar the colours, the title and the comment — `maximum-scale=1`,
`user-scalable=no`, `viewport-fit=cover`, `overflow: hidden`,
`overscroll-behavior: none`, `user-select: none`, `touch-action: none` — and
`chrome.test.ts` asserts each token in both files.

**How to play.** SIEGE opened on a board of empty sockets, a river of lava and
the words "HOLD THE FORGE", and nothing anywhere said that the sums at the
bottom are how you pay for the guns. A child who does not work that out in the
first thirty seconds watches the wave walk into the core and decides the game is
broken.

## Blast radius

**Areas touched:** dynawalla (one game pack)

- [x] Every touched path is covered by a `ci.yml` area filter.
- [x] **Pack back-compat.** `pack.json` untouched: same id, same version, same
      capabilities (`items`, `items.reveal`, `haptics`), same 4 `covers.skills`.
      No catalog entry changed, no published zip changed in place.
- [x] **Native integrity.** N/A — no Rust, no `src-tauri`, no manifests.
- [x] **Strings.** N/A — no `corpan-app` locale keys. The new strings are the
      pack's own instruction copy, written for a 7–11 year old.
- [x] **Curriculum.** N/A — no skill ids, generators or schemas touched.
- [x] **Secrets.** None. `gitleaks` output below.

## Proof

```
$ cd dynawalla/games/siege && for i in 1 2 3 4 5; do npm test; done
run 1: # pass 29 # fail 0
run 2: # pass 29 # fail 0
run 3: # pass 29 # fail 0
run 4: # pass 29 # fail 0
run 5: # pass 29 # fail 0

$ npm run tsc
> tsc --noEmit && tsc --noEmit -p tsconfig.test.json
(no output — 0 errors, both projects)

$ npm run build
✓ 28 modules transformed.
dist/index.html                  1.24 kB │ gzip:  0.64 kB
dist/assets/index-B9gfJ_kq.css  13.07 kB │ gzip:  3.69 kB
dist/assets/index-CsYwaTFB.js   78.04 kB │ gzip: 27.04 kB
✓ built in 279ms

$ cd ../../packs && node build.mjs siege
dynawalla.siege@0.1.0 — SIEGE
  entry        pack.html
  host         0.1.0 … <1.0.0
  sdk          1.0.0 (this SDK is 1.0.0)
  capabilities items, items.reveal, haptics
  covers       4 skills, grades 1–4
  on disk      4 files, 95456 bytes

1 pack(s) built, checked and staged into src-tauri/packs/

$ gitleaks detect --no-banner --redact --log-opts="origin/main..HEAD"
INF 1 commits scanned.
INF scanned ~28694 bytes (28.69 KB) in 45.9ms
INF no leaks found

$ git diff --name-only --diff-filter=D origin/main...HEAD
(empty)

$ git diff --name-only origin/main...HEAD
dynawalla/games/siege/README.md
dynawalla/games/siege/index.html
dynawalla/games/siege/pack.html
dynawalla/games/siege/src/mount.ts
dynawalla/games/siege/src/render/draw.ts
dynawalla/games/siege/src/ui/chrome.test.ts
dynawalla/games/siege/src/ui/chrome.ts
dynawalla/games/siege/src/ui/hud.ts
dynawalla/games/siege/src/ui/styles.css
```

### Every fix fails the test when removed

Each was reverted on its own and the suite re-run:

```
### FIX 1 REMOVED (CORNER_CLEAR = 0) ###
not ok 1 - the status bar's contents never sit under the host's two controls
not ok 2 - the clearance is exactly one control and its margin, never a band
# pass 7 / # fail 2

### FIX 2 REMOVED (the side env() rules dropped from styles.css) ###
not ok 5 - the stylesheet honours all four edges, not only the two obvious ones
# pass 8 / # fail 1

### FIX 3 REMOVED (board fitted to the element, not the safe box) ###
not ok 8 - the board fit obeys the safe box by construction, not by luck
# pass 9 / # fail 1

### FIX 4 REMOVED (gesture guards) ###
not ok 9 - both entries block the double-tap that scales and pans the host document
# pass 8 / # fail 1

### RESTORED ###
# pass 29 / # fail 0
```

**On fix 3, honestly.** The first version of that test asserted the board fit
across the five viewports and **passed either way** — on every shape the fleet
has today the board is height-limited and the insets are near enough symmetric,
so honouring the safe box changes the fit by nothing. A clearance test that
passes either way is worthless, so it was replaced with one that hands the fit a
deliberately hostile safe box and asserts the result differs from ignoring it.
The test now says in as many words that on today's shapes it is a no-op kept
true by construction, because it is one screen shape — a split view, a foldable,
a future cutout — away from being a socket a child taps and cannot build.

Viewports swept: 320×568, 390×844, 768×1024, 1024×768, 844×390, with no insets,
with the portrait cutout, and — for landscape — both the symmetric case and the
one-side case WebKit actually reports.

### DIFF
--- a/dynawalla/games/siege/README.md
+++ b/dynawalla/games/siege/README.md
@@ -108,7 +108,7 @@ src/core/           rng · easing · Clock (hitstop, slow-mo, fast-forward) · C
 src/audio/          the whole sound palette
 src/game/           constants (all balance) · path · board · waves · state + step
 src/render/         bake (static board) · particles · draw
-src/ui/             hud (DOM) · styles
+src/ui/             hud (DOM) · styles · chrome (safe area, the host's two corners)
 src/mount.ts        the controller: loop, input, sim events → light and noise
 ```
 
@@ -124,6 +124,32 @@ full-width overcharge bar. **Desktop**: a live tower palette with cost against d
 then click pads to place, with the range previewing under the cursor. `1`–`4` answer, `Q`/`W`/`E`
 arm bolt/mortar/chain, `U` upgrade, `Space` overcharges or calls the wave in early for bonus
 embers, `F` fast-forwards, `Esc` closes.
+
+## The room the host leaves us
+
+SIEGE was one of the seven games that already read `env(safe-area-inset-*)`, and it was the
+half-fix that usually is: `.sg-top` honoured `--top`, `.sg-anvil` honoured `--bottom`, and
+**neither side was touched at all**. `viewport-fit=cover` opts a document into the rounded corners
+on every edge, and held sideways the cutout is a *side* inset — which is where the ember count and
+the sound switch were. All four edges are honoured now, on the status bar, the console, the
+overlays and the banners, and the square board is fitted inside the safe box because a socket
+under a rounded corner is a tower a child taps and cannot build.
+
+The host also paints an exit control over the top-left 44px corner and a how-to-play control over
+the top-right one, on top of this pack. The status bar ran the full width of exactly that row. Its
+*contents* now start after the first control and stop before the second — the bar keeps its
+height, the board keeps its size, and the forge still reaches the glass. Reserving the whole strip
+instead was tried in a sibling game and cost 12% of a small phone's height.
+
+Paying 108px for those two corners meant the three switches — call wave, speed, sound — had to
+leave the bar. They are in the anvil's head row now. That bar was over-full on a 320px phone
+anyway: `overflow: hidden` was quietly cutting those same three switches off the right-hand edge,
+so nobody on a small phone could reach them at all.
+
+`src/ui/chrome.ts` holds every number involved, derived from the shared host constants, and
+`chrome.test.ts` asserts it at five viewports in both rotations. `pack.html` also gained the
+`maximum-scale=1` and `touch-action: none` guards that every other pack of the twenty-seven
+already had; without them a double tap inside SIEGE could scale and pan the host document.
 
 ## Swapping in the real host
 
--- a/dynawalla/games/siege/index.html
+++ b/dynawalla/games/siege/index.html
@@ -4,8 +4,9 @@
     <meta charset="utf-8" />
     <meta
       name="viewport"
-      content="width=device-width, initial-scale=1, viewport-fit=cover, user-scalable=no"
+      content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no, viewport-fit=cover"
     />
+    <meta name="color-scheme" content="dark" />
     <meta name="theme-color" content="#0a0608" />
     <title>SIEGE</title>
     <link
@@ -18,9 +19,13 @@
         margin: 0;
         padding: 0;
         height: 100%;
+        overflow: hidden;
         background: #0a0608;
         overscroll-behavior: none;
         -webkit-tap-highlight-color: transparent;
+        -webkit-user-select: none;
+        user-select: none;
+        touch-action: none;
       }
       #app {
         position: fixed;
--- a/dynawalla/games/siege/pack.html
+++ b/dynawalla/games/siege/pack.html
@@ -4,8 +4,9 @@
     <meta charset="utf-8" />
     <meta
       name="viewport"
-      content="width=device-width, initial-scale=1, viewport-fit=cover, user-scalable=no"
+      content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no, viewport-fit=cover"
     />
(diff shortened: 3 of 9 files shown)


skyl:
Amended and force-pushed after a self-review pass.

**What changed.** Moving the three switches into the anvil's head row is safe in the portrait and wide layouts, but the `@media (max-height: 619px)` branch turns the console into `1fr 116px` — and 568 is under 619, so a **320×568 phone takes that branch too**. Its left column is about 180px. `ANVIL` + `+0` + the speed and sound chips fit; add the "call the wave in early" chip that appears during an intermission and it does not, and a `nowrap` flex row spills rather than reflowing.

So `.sg-head` gains `flex-wrap: wrap` and `min-width: 0`, and the short-viewport branch compacts the chips (`padding: 5px 7px`, `font-size: 10px`, `min-height: 28px`, tighter gaps). The row now wraps to a second line for the few seconds the call chip is showing, and only then.

Re-ran every gate on the amended commit:

```
$ for i in 1 2 3 4 5; do npm test; done
run 1: # pass 29 # fail 0
run 2: # pass 29 # fail 0
run 3: # pass 29 # fail 0
run 4: # pass 29 # fail 0
run 5: # pass 29 # fail 0

$ npm run tsc
> tsc --noEmit && tsc --noEmit -p tsconfig.test.json
(no output — 0 errors, both projects)

$ npm run build
dist/assets/index-CJ9RfpPU.css  13.20 kB │ gzip:  3.72 kB
dist/assets/index-BhekiJ5y.js   78.04 kB │ gzip: 27.04 kB
✓ built in 158ms

$ cd ../../packs && node build.mjs siege
  covers       4 skills, grades 1–4
  on disk      4 files, 95577 bytes

1 pack(s) built, checked and staged into src-tauri/packs/

$ gitleaks detect --no-banner --redact --log-opts="origin/main..HEAD"
INF 1 commits scanned.
INF scanned ~29154 bytes (29.15 KB) in 40.1ms
INF no leaks found
```

PR #659, opened 2026-07-29
dynawalla/host: the difficulty wire — next() has been ignoring its argument in production

skyl:
Ten games have been calling `host.next({ difficulty })` for months. The shared adapter's signature was `next(): Question`, wired `next: take`, and `take()` ignored everything passed to it. Every one of those arguments was discarded in production. The games were talking into a dead wire.

This PR is the **wire**, not the controller. It does not decide anything. The controller lives in `packs/shared/game-pacing/` and is somebody else's file.

## What was broken

| | |
|---|---|
| `GameHost.next()` | took no arguments; `HostClient.nextItem` had no difficulty on the wire at all |
| `Question.difficulty` | `item.level / 8` — a read-back, and of the **wrong number**. `level` is the level *within* a skill (0..3 on the shipped graph), not a ladder position, so the hardest question in the product read `0.375`. colossus and siege branch on 0..1 and could not tell the top of the ladder from the middle |
| `report()` | ended in a literal `void correct` — 25 of 27 games know locally whether the child was right, and the adapter threw it away |
| the pool | 32–64 item FIFO between any decision and the child, so a change landed **34 questions** later |
| `setDifficulty` / `raiseFloor` | trebuchet and siege have feature-detected these since they shipped; the adapter did not implement them |

## The wire

```ts
next(request?: { domain?: string; difficulty?: number; maxDifficulty?: number }): Question
flush(): void
setDifficulty(d: number): void
raiseFloor(d: number): void
recentOutcomes(): readonly Outcome[]
```

The request shape is the one those ten call sites **already produce**, so adoption is free. `difficulty` reaches `items.next` as a 0..1 position on the host's whole ladder, selects the rung, and moves the host's own ladder position — so a pack that stops driving resumes from where it left the child rather than being pulled back up by the ladder's own climb.

`Item.difficulty` (optional, so an older host is not a breaking change) carries the ladder ordinate the host **actually used**, which is how a pack can see a clamp.

`Outcome` keeps both verdicts: `correct` is the host's and is the record; `claimed` is the game's, available a round trip sooner, and the only one there is when the round trip never comes back (`judged: false`).

## Two scales, one field

Both are in production and neither can be removed: six games send an integer 1..10 ladder index, three send a 0..1 fraction, runner sends an unrounded 0..12. `toUnit` states the rule — `< 1` is a fraction, `>= 1` is `(v-1)/9` — leaving exactly one ambiguous value, `1`. It is read as the **bottom**, because five of the six ladder games send it on their opening question and serving the hardest content in the product to a child who has just started is the failure this exists to prevent. It says so out loud, once, when it sees one. Out of range clamps loudly. Not a number is ignored loudly.

## When a request cannot be satisfied

The scale is relative to the host's own ladder, so `0` always lands on the easiest rung that exists — today two-digit + two-digit, because the curriculum floor is still being authored. That clamp is **announced**, naming the rung it stopped at:

```
[pack] difficulty 0.00 was requested and the easiest rung this host has is
"arith.add.two-digit" at 0.00 — the curriculum has nothing below it, so this
is as easy as questions get today
```

When rungs below it land, the same request follows them down and no pack is rebuilt.

## Flush, measured

`34 → 2` questions between a difficulty change and a question that reflects it, asserted in `index.test.ts` rather than claimed. Flush never empties the pool — an empty pool is not a pause, it is a question with no id whose answer gets dropped — so it keeps the eight closest questions as a reserve, plus anything `focus` asked for.

## Verification

- **`game-host` had no `package.json`**, so CI's `dynawalla/packs/shared/*/package.json` glob ran no tests and no typecheck against the module 27 packs mount on. It has one now.
- **16 fix-deletions**, each confirmed to turn the suite red, with the failure message recorded.
- **All three suites run 5×**, clean: game-host 17, sdk 54, app 225.
- **15 games** installed, typechecked and tested against the new contract: 1084 tests, 0 failures. A probe outside `games/` confirms the widened `GameHost` structurally satisfies all 15 local `Host` contracts **without** the `as unknown as` cast that would otherwise hide a mismatch (validated against a negative control).
- **No Tauri boundary.** The item wire is a `MessagePort` between two documents in one WebView; `src-tauri/src/packs/mod.rs` never sees an `Item`. Checked by grep, not assumed, because a Rust `models.rs` silently dropping unknown serde fields has bitten this repo before.

## Touches nothing under `games/`, `game-chrome/`, `curriculum/`

Eight agents are working in those. `git status` on this branch is clean outside `packs/shared/game-host/`, `packs/sdk/src/` and `dynawalla-app/src/packs/`.

## Behaviour changes worth a runtime look

`siege`'s `raiseFloor` (0..0.72) and `trebuchet`'s `setDifficulty` (capped at 0.95) were no-ops and are now live calls. Both stay clear of the ambiguous `1`. `polarity` reaches exactly `1` after ~7.5 minutes of survival and means *hardest*; it will be read as easiest and will log the notice. That is the documented cost of the one ambiguous value, and the log line is how its owner will find it.

### DIFF
--- a/dynawalla/dynawalla-app/src/packs/bridge.test.ts
+++ b/dynawalla/dynawalla-app/src/packs/bridge.test.ts
@@ -343,3 +343,36 @@ test("the pack id the host is told is the host's, never the pack's", async () =>
   })
   assert.deepEqual(calls[0]?.input, { packId: "abacus.tower", skillId: "add.1" })
 })
+
+test("a difficulty request crosses the boundary, clamped rather than refused", async () => {
+  const calls: Calls = []
+  const bridge = createBridge({ packId: "p", granted: ["items"], services: services(calls) })
+
+  await bridge.handle({
+    id: 1,
+    method: "items.next",
+    params: { difficulty: 0.25, maxDifficulty: 0.75 },
+  })
+  assert.deepEqual(calls[0]?.input, {
+    packId: "p",
+    difficulty: 0.25,
+    maxDifficulty: 0.75,
+  })
+
+  // Out of range is a bug in a pack, and refusing the question would turn it
+  // into a blank screen in a child's game. Clamped, like every other number
+  // this boundary takes, and announced on the pack's side where an author can
+  // act on it.
+  const clamped = await bridge.handle({
+    id: 2,
+    method: "items.next",
+    params: { difficulty: 4, maxDifficulty: -3 },
+  })
+  assert.equal(clamped?.ok, true)
+  assert.deepEqual(calls[1]?.input, { packId: "p", difficulty: 1, maxDifficulty: 0 })
+
+  // Not a number at all is absent, not zero: a pack that sends a string must
+  // not thereby pin every question to the easiest rung on the ladder.
+  await bridge.handle({ id: 3, method: "items.next", params: { difficulty: "hard" } })
+  assert.deepEqual(calls[2]?.input, { packId: "p" })
+})
--- a/dynawalla/dynawalla-app/src/packs/bridge.ts
+++ b/dynawalla/dynawalla-app/src/packs/bridge.ts
@@ -40,6 +40,7 @@ import {
   parseRequest,
   permits,
   stringParam,
+  unitParam,
 } from "../../../packs/sdk/src/index.ts"
 
 /** A pack's own key-value store. Small on purpose: it is not a database. */
@@ -63,7 +64,14 @@ const SOUND_CUES: readonly SoundCue[] = ["tick", "seat", "settle", "refuse", "ar
  * only one file.
  */
 export type HostServices = {
-  nextItem(input: { packId: string; skillId?: string }): Promise<Item | null>
+  nextItem(input: {
+    packId: string
+    skillId?: string
+    /** 0..1 across the host's whole ladder. A request, not an instruction. */
+    difficulty?: number
+    /** 0..1 ceiling. The stream never goes above it. */
+    maxDifficulty?: number
+  }): Promise<Item | null>
   /** Records the attempt and judges it. One call, in that order. */
   judge(input: {
     packId: string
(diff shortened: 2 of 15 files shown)


skyl:
### Second commit: main was already red, and the reason is in this PR's blast radius

CI failed on the merge tree with two `items.test.ts` failures. They reproduce on `origin/main` **without this branch** — `dynawalla-curriculum: the ladder now starts at 0 + 1` (#655) landed and main has been red since.

The underlying defect is not a test problem. `items.ts` read its operands out of slots it named — `top`/`bottom` — and `slotText(undefined)` returns `""`. #655 added a second active family, `gen.arith.number-facts`, whose slots are `first`/`second`:

```
served: " + "   digits 1     ← every question on the six easiest rungs
```

No numbers, no throw, no log. Same shape on the operator: both active families define their own `PROMPT_KEY_SUB` (the curriculum's `index.ts` re-exports them by name *because* they collide), and this file compared against column-op's alone — so every number-facts subtraction was drawn with a plus sign.

I fixed it here rather than handing it off because **this PR is what makes it reachable**: `difficulty: 0` did not exist before today, and it routes children straight onto those rungs.

After the fix, on the post-#655 tree:

```
0     "0 + 3"          dw.add.facts.add-within-ten          ord 0.00
0.05  "3 − 2"          dw.add.facts.subtract-within-ten     ord 0.06
0.15  "5 − 1"          dw.add.facts.subtract-within-ten     ord 0.14
0.5   "1847 + 5022"    dw.add.column.add-no-regroup         ord 0.51
1     "7003 − 5494"    dw.add.regroup.subtract-across-zero  ord 1.00
```

36 rungs, single-digit facts to four-digit borrow-across-zero, addressed 0..1 end to end. That is the founder's ask working.

Fixed **generally**, not by adding `first`/`second` to a list: operands come from the slots a family declares in declaration order when it does not use the named pair; the operator comes from the last key segment, which every family agrees on. Seven more families are registered and inactive; the next one to be activated is now covered by existing rather than by somebody remembering to edit this file. An empty operand is **refused**, loudly, naming the skill, family and slot keys — a blank question is worse than no question.

Two pre-existing tests were passing wrongly rather than failing, both fixture assumptions the new single-digit content broke: one used `"0"` as its stand-in for a wrong answer (zero is a fine answer to `3 − 3`, so it was climbing the ladder while asserting it had stepped down), and the leak detector scanned the whole item for the canonical value, which `0 + 3` carries on its face.

New guard test walks **every** rung and asserts two numeric operands, a readable prompt, and an operator that agrees with the skill. Fix-deletion confirmed:

```
AssertionError: dw.add.facts.subtract-within-ten was drawn as an addition
[packs] dw.add.facts.add-within-ten (gen.arith.number-facts) drew a prompt with a
        missing operand: slots are [first, second] and this read ["", ""]
```

Verified green on **both** trees — this branch, and this branch's files applied over `origin/main` — 227 tests, lint and tsc clean, app suite 5×.

PR #670, opened 2026-07-29
feat(dynawalla): the harness applies the system

skyl:
## What

Component half of the harness pass. **Stacked on #669** — base is
`dw-premium-tokens`, not `main`. Merge #669 first and this retargets to main.

## Why

Three defects that only a sweep of every screen at every width in both themes
actually finds.

**Every segmented control in the app was drawn inverted.** The SELECTED option
got the recessed ground and the unselected ones the raised surface — so the
chosen segment read as a hole and the unchosen ones as pressable keys. Six
controls on Settings, two on Parents, and the catalogue's subject chips. Now a
recessed track with a raised thumb.

**The strapwork was drawn twice on every screen** — under the lintel and above
the tab bar. Once now.

**Four competing measures on desktop.** Lintel, surface, courses and tab bar each
began at a different x, and Parents used about a fifth of a 1440 screen. One
geometry now: wordmark, first row, first card and first tab share an origin at
every width; only prose stays narrower. Parents goes two-column above 64rem.

Also: the tab bar is a tab bar — casts upward, content scrolls under it, targets
clear 44px, clears the home indicator and the Android gesture bar. And cards
answer a touch, which they did not — the press state was hover-only, so on a
phone a child's tap was acknowledged by nothing until the next screen appeared.

## What the verifiers caught

Four independent lenses tried to refute this work. They disproved **two claims
made in it**:

- *"a choice row is now the same 64px object as a fact row"* — false at every
  width, measured. Fixed.
- *"the wordmark, the writing and the tabs share one geometry"* — three x-origins
  had become two, not one. Fixed.

Plus: the tab bar **clipped its own labels** at the Largest text size the app
ships ("Prog…", "Profi…", "Setti…") on every phone width — now capped by
container query, zero clipped labels across 6 widths × 3 text sizes. And a
`:focus-within` ring appeared **on touch**; now `:has(:focus-visible)`.

## Blast radius

- [x] Chrome and surfaces only. No game, no pack, no native code, no launch path.
- [x] `surfaces.test.ts` untouched and passing — it is load-bearing: two
      destinations rendered empty for weeks with a green suite before it existed.
- [x] No React `style` prop anywhere; `style-src 'self'` silently discards them
      on device. The test asserting this passes, and its glob was re-verified.
- [x] `dynawalla/games` and `dynawalla/packs` untouched — verified empty diff.
- [ ] **Not done, deliberately:** the pass sheet is unreachable on device because
      billing is unwired — a wiring change, not a design one. And a game's key
      art keeps its colours in both themes: *one warm point* governs the chrome,
      not the picture inside a card.

## Proof

```
npm run tsc     clean
npm test        tests 243 / pass 243 / fail 0   (238 on the base)
npm run lint    clean
npm run build   ✓ built in 151ms
```

120 screenshots re-captured after every change and reviewed at 390 / 834 / 1440
in both themes.

### DIFF
--- a/dynawalla/docs/HARNESS_FEEDBACK.md
+++ b/dynawalla/docs/HARNESS_FEEDBACK.md
@@ -73,13 +73,91 @@ pressing hard on a card got a text-selection handle over the label. `user-select
 none` on `body`, with `input`, `textarea` and `[contenteditable]` opting back in
 so a parent can still edit a name they typed.
 
+### "Erase everything" was coral — 2026-07-29
+
+Was: `--dw-strike` came off the copper ramp, a leftover from the pre-purple
+palette, and a bare coral phrase floated in white space with nothing around it
+to say it was the most consequential control in the app.
+
+Fixed, and the decision was taken rather than drifted into. The strike ramp is
+now **rose**, and its hue is the arc's own rose end (`--color-arc-11`, hue 348°)
+— danger drawn in a colour the app already contains rather than a fifth family.
+It is three tokens, not one: `--dw-strike` (the word), `--dw-strike-line` (the
+frame it is bounded by) and `--dw-strike-ground` (the wash it fills with once
+armed), so the control is a bounded plate that visibly changes state before the
+press that actually erases. Light measures 6.39:1, up from copper's 4.45:1.
+
+The plate's resting border is `--dw-line-strong`, not `--dw-line`: at
+`--dw-line` it was the same rgb as the hairline between two ordinary rows, so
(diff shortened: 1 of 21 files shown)


skyl:
Closing this stale component-half PR. The foundation in #669 landed, but this branch now conflicts with the current Catalog, stylesheet, and capture tooling and predates substantial subsequent Dynawalla changes. A 21-file visual sweep should not be conflict-resolved blindly; any still-useful pieces should return as small, fresh main-based PRs with current screenshot evidence. There is no safe squash delta to merge from this branch as-is.

PR #694, opened 2026-07-29
fix(dynawalla/games/polarity): eight characters was never measured, and a rung it cannot draw is a Bearer that asks nothing

skyl:
## What

Make a ten-character answer drawable on a POLARITY orb, cap the stream at what
the pack can draw instead of declining item by item, and promote the last two
multiplication rows — which puts `48,826 × 82,726` in front of a child.

## The real width constraint, which is not the atlas

PR #683 recorded that POLARITY's numeral cell "holds eight characters". It does
not hold eight; it *permits* eight, and the reason written next to the constant
was false:

```ts
/**
 * … Eight leaves a character of headroom and
 * still fits the cell without squeezing.
 */
export const LABEL_MAX_CHARS = 8;
```

`atlas.ts` has always **fitted** rather than clipped — `if (w > maxW) g.scale(maxW / w, 1)`
— so nothing overflows a cell and nothing ever refuses for want of texels. Eight
characters at the 76-unit type size measure about **365** design units against a
**232**-unit box, so the widest numeral the old constant allowed was already
squeezed to roughly **0.64** of its natural width. Nothing measured that and
nothing enforced it. No shipped item ever reached it either: the widest answer
the ladder emitted was six characters.

The texture is not the constraint either, though it is genuinely out of room
sideways: the atlas was 2048 × 768 and 2048 is the WebGL-guaranteed
`MAX_TEXTURE_SIZE` floor. But a cell is a *sampling resolution*, and the label's
size on screen has nothing to do with it.

**The constraint is the orb's lane.** Four orbs share `(HALF_W − 12) × 2 = 76` of
the hundred-unit playfield, so each gets 19 units; the label drawn over one is
`orbR × 1.35 × LABEL_ASPECT × boost` = 17.2 units wide on a narrow phone. Two
labels that overlap are worse than one that is condensed, so that 19 units is the
whole budget, and the on-screen condensation of an *n*-character numeral works out
to exactly

```
advance / em  =  (LABEL_CELL_H · LABEL_ASPECT − margin) / (LABEL_EM · n)
```

— font-independent, because anything wider than the box is squeezed to fill it
exactly. Eight characters is simply where that ratio, at the old aspect and
margin, happened to land on 0.38 em.

## What changed

**The budget is derived, not typed.** `LABEL_MIN_ADVANCE_CAPS = 0.5` — a digit may
not be narrower than half its own cap height, which is where condensed tabular
faces bottom out — and `LABEL_MAX_CHARS` falls out of the geometry at **ten**,
with less than one percent of margin. `labels.test.ts` asserts both that ten fits
and that eleven does not, so a later change to the cell, the margin or the aspect
moves the number rather than silently keeping it.

**The width it needed came from geometry, not from lowering a bar:**

- `LABEL_ASPECT` 2 → 2.25, and the grid 8 × 6 → **7 × 7** so the texture *shrinks*
  sideways: 2016 × 896 against the 2048 ceiling, at the same 48 claimable cells.
  A new guard shouts if `cellPx * LABEL_ASPECT` is ever not a whole number of
  texels — the shader addresses a tile as a fraction of `uGrid`, so a rounded cell
  width is a numeral sampling a sliver of its neighbour, and it reads as a font bug.
- `LABEL_CELL_MARGIN` 24 → **20**. The number it has to clear is the baked contrast
  rim, which puts ink 6.5 design units outside the outline (`lineWidth 13`); 24 was
  a round number nothing derived, and the four units are the difference between ten
  characters and nine.
- `ORB_SPREAD` `(HALF_W − 12) × 2` → `(HALF_W − 8) × 2`. The outermost orb moves from
  28.5 to 31.5 and its numeral's far edge lands at 40 of the 50 units the field has,
  nowhere near the weave clamp at `HALF_W − r`.

**Type size does not move.** `LABEL_EM / LABEL_CELL_H` is unchanged, so every numeral
is drawn at the height it always was and only the advance changes with length. Across
five viewports the test computes cap height from the constants and holds it to the
arcade canon's hard gate for numerals on moving objects — 22 rpx "at the moment of
decision" — which it clears at 29–39 rpx everywhere. `games/beam`'s 15 css px
candidate floor is cleared in portrait (17–20 px); the landscape phone reads 14.2 px
and always has, which this PR does not change and does not claim to.

Short numerals get *better*, not worse: up to five characters they are now drawn
undistorted where they used to start squeezing at six.

**`maxDifficulty`, which is the durable fix.** `next({ difficulty, maxDifficulty })`
has existed on `GameHost` since #659 and POLARITY passed neither. It does now, and
the mechanism does not need to know anything about the curriculum: when
`drawAskable` meets an item it cannot print, `capBelow` sets the ceiling to that
item's own reported ladder ordinate minus a thousandth — which is strictly inside
one rung on any ladder shorter than a thousand rungs, so the host's `floor(cap × span)`
excludes exactly that rung — and calls `host.flush?.()`, because the pool holds up to
sixty-four questions all drawn from the rung that was just ruled out.

Monotone downwards, never raised, and not reset by `startRun`: what POLARITY can draw
is a property of POLARITY, so re-learning it would cost a starved Bearer every restart.

**A third defect, at the same call site.** `game-host`'s `toUnit` reads a value below 1
as a fraction and 1..10 as a ladder index, and resolves the one value both scales claim
— `1` — as the **BOTTOM**, because six other games send it on their opening question.
POLARITY sent `clamp(0.14 + stratum * 0.06, 0, 1)`, which pins at exactly 1 from
stratum 15 — about seven and a half minutes of good play — and meant the **top**. So a
child who had climbed fifteen strata was dropped to the easiest content in the product
and held there for the rest of the run. `game-host` names this game in the comment on
that rule. It sends `1 + unit × 9` now, on both fields.

The stub host learned the same scale and the same ceiling, transcribed with a comment:
a dev harness that reads the request differently from the shipping host is how the
blank-orb defect survived a whole build.

## Measured, 60 seeds a level, before and after

Answers wider than the budget, over the whole promoted ladder (66 rungs). Only the
three levels that were ever wide are shown; nothing else on the ladder exceeds six
characters.

| level | seen | too wide @ 8 | too wide @ 10 | **declined** @ 10 | widest |
|---|---|---|---|---|---|
| `dw.mul.scale.times-power-of-ten L2` | 60 | 17/60 | **0/60** | **0/60** | `635,560,000` — 9 |
| `dw.mul.multidigit.long-multiplication L1` | 60 | 52/60 | **0/60** | **0/60** | `105,342,552` — 9 |
| `dw.mul.multidigit.long-multiplication L2` | 60 | **60/60** | **0/60** | **0/60** | `1,234,138,608` — 10 |

`declined` is `orbValues()` returning `null` — the pack's own refusal, including
distractors. Extended to **1,000 seeds a level**: still 0 declined, widest answer
`1,234,138,608` and widest distractor `1,234,138,609`, both ten. Eleven is not
reachable by construction — five digits by five digits is at most ten characters,
and the mal-rules are smaller than the product, not larger.

#683 measured 21/60, 50/60 and 60/60 on the same levels; the small differences are
the seed derivation (`seedFrom` here, raw indices there) and not the content.

## `48,826 × 82,726`, end to end

Through the real `createItemService`, at `difficulty: 1`, forty items — every one
drawn by `orbValues`, every orb given a real tile by `LabelBook`, every canonical
answer fed back through `judge`:

```
rungs: 66
  dw.mul.multidigit.long-multiplication L2  "42539 × 12532" = 533098748
      orbs [533098748, 533098749, 533098747, 553007]   correct=true
  dw.mul.multidigit.long-multiplication L2  "31457 × 55613" = 1749418141
      orbs [1749418141, 629140, 1749418140, 1749418142] correct=true
  dw.mul.multidigit.long-multiplication L2  "75513 × 63536" = 4797793968
      orbs [4797793968, 1736799, 4797793969, 4797793967] correct=true
drawn 40/40   judged correct 40/40

48,826 × 82,726 = 4039179676 -> "4039179676" (10 chars)
a real ten-character item off the top rung: "36536 × 69356" = 2533990816 at difficulty 1
```

`48,826 × 82,726` is exactly the shape of `long-multiplication` L2 — five digits by
five — its product is ten characters, and ten characters is what POLARITY prints.

## Rows promoted

- `dw.mul.scale.times-power-of-ten` (draft → active)
- `dw.mul.multidigit.long-multiplication` (draft → active)

A `status` flip and nothing else: no generator, level table or parameter changed.
`NUMERAL_WIDTH_BLOCKED_LEVELS` is now empty and `SHIPPED_NUMERAL_MAX_CHARS` is 10.
The empty list stays rather than being deleted, because `render/prompts.test.ts`
asserts it in **both** directions — the day a row reaches eleven characters it is
named by a failing build rather than rediscovered by promoting rows and watching a
pack's sweep go red, which is how the first one was found.

## Removed-fix verification

Every test was checked by deleting the fix it covers.

Narrow `ORB_SPREAD` back to `(HALF_W − 12) × 2`:

```
✖ a wide numeral still fits its own orb's lane
  a numeral is 19.4 wide in a 19.0 lane — orbs would collide
```

Type `LABEL_MAX_CHARS = 8` back in (three failures, the first over the real,
promoted ladder — the symptom #683 saw in CI):

```
✖ not one orb the shipping curriculum can put on the field comes out blank
  dw.mul.scale.times-power-of-ten L2 produced an item with no drawable answer
✖ every value the ladder emits survives a round trip through a printed numeral
✖ the longest numeral is derived from the lane, and is not a number somebody typed
  9 characters would also have fitted — the cap is not the geometry
```

Shrink `LABEL_EM` (the constant that sets type size):

```
✖ the longest numeral clears the canon's cap-height gate on every viewport
  a phone, landscape: cap height is 18.8 rpx, under the canon's 22
```

Send the raw fraction again instead of `ladderScale` (four failures):

```
✖ the top of the ladder is asked for as the TOP of the ladder
  the hardest content in the product was asked for as the easiest
  1 !== 10
```

Delete `capBelow` — **the soft-lock, reproduced**:

```
✖ a rung POLARITY cannot draw is asked once, not forever
  six Seal Bearers arrived and not one of them asked a question — the child at
  the top of the ladder was served silence
✖ the ceiling is learned once and costs one Bearer, not one per Bearer
  a later Bearer had to rediscover the ceiling
✖ the ceiling only ever falls, and the pool is flushed when it does
✖ an item with an answer this game cannot print is declined, not shown unlabelled
```

Drop the `host.flush?.()`:

```
✖ the ceiling only ever falls, and the pool is flushed when it does
  a ceiling landed sixty-four questions late
```

Drop the monotone guard inside `capBelow` — exercised by the reserve a real flush
leaves, which `game-host` documents as still answerable above a ceiling "when
nothing else does":

```
✖ a ceiling is never raised by the stale question that outlived it
  the run stopped asking questions altogether
```

Cap on every ask rather than only on a refusal:

```
✖ a rung POLARITY cannot draw is asked once, not forever
  it fell back to 6 characters when it can print 10
✖ a host with no ceiling to give is not made worse by asking for one
  a ladder this game can draw entirely was still capped
```

Revert `SHIPPED_NUMERAL_MAX_CHARS` to 8, and separately to 9:

```
✖ the levels whose answers are too wide to print are exactly the ones promotionBlockers.ts names
  + ['dw.mul.multidigit.long-multiplication L1',
  +  'dw.mul.multidigit.long-multiplication L2',
  +  'dw.mul.scale.times-power-of-ten L2']
  - []

✖ the levels whose answers are too wide to print are exactly the ones promotionBlockers.ts names
  + ['dw.mul.multidigit.long-multiplication L2']
  - []
```

## Suites, five runs each

```
polarity   ×5:  ℹ tests 60  ℹ pass 60  ℹ fail 0
curriculum ×5:  ℹ tests 208 ℹ pass 208 ℹ fail 0   |  npm run check: OK
app        ×5:  ℹ tests 292 ℹ pass 292 ℹ fail 0
```

**Every game in the tree**, because the ladder moved under all of them:

```
arena 65   balance 68   beam 99    claim 78    coil 109   colossus 81
counterweight 136  forge 79   foundry 93  guilty 36   horde 29   lattice 94
merge-idle 175  merge 73   mosaic 79   polarity 60  pulse 129  rhythm 70
runner 73  serpent 38  siege 29   skyledger 69  slice 114  stack 36
street 99  trebuchet 166  truedraw 154        — 0 failures
```

`packs/sdk`: 88 pass. Typecheck clean on polarity (both projects), curriculum and
the app; `npm run lint` clean; `build:pack` clean.

## Blast radius

**Areas touched:** dynawalla (`games/polarity/`, `packs/shared/curriculum/`)

- [x] Every touched path is covered by a `ci.yml` area filter — `dynawalla/games/`
      by `dynawalla_packs`, `dynawalla/packs/shared/curriculum/` by
      `dynawalla_curriculum` and `dynawalla_packs`. Observed on this PR:
      `dynawalla-app`, `dynawalla-curriculum` (ubuntu **and** macos) and
      `dynawalla-packs` all ran, and the app's 292 tests — which sweep the ladder —
      are green in CI as well as five times locally.
- [x] **Outside the stated file list, deliberately and minimally:**
      `packs/shared/curriculum/src/snapshots/generators.json`. CG-16 requires a
      committed determinism hash per active (skill, level), so the six new hashes
      are a mandatory consequence of the two `status` flips — `npm run check` is
      FAILED without them. Produced by `npm run snapshots:update`, no generator
      touched. Nothing under `games/arena|balance|beam|colossus|counterweight|merge-idle|serpent|siege|skyledger|slice|stack|street`,
      `packs/sdk/` or `dynawalla-app/` was modified.
- [x] **Pack back-compat.** No published pack zip changed, no catalog entry touched,
      no `minAppVersion` moved. `Item`, `Judgement` and the `packs/sdk` protocol are
      unchanged — `ItemRequest.maxDifficulty` and `GameHost.flush` both already
      existed and are used here for the first time by this game.
      `packs/shared/game-host` is untouched. POLARITY's local `Host` type gained
      `Ask.maxDifficulty` and an **optional** `flush?()`, which keeps it structurally
      assignable from `GameHost` and feature-detected exactly as `trebuchet` and
      `siege` do with theirs.
- [x] **Native integrity.** N/A — no Rust touched.
- [x] **Strings.** N/A — no locale strings. The two promoted rows' `locKey`s already
      exist and are unchanged.
- [x] **Curriculum.** No skill id renamed or reused. No grade change. No generator,
      parameter or difficulty coefficient changed. Every operand and answer stays an
      exact `Rational`. Two rows promoted; both have passing generator bindings and
      `npm run check` is OK.
- [x] **Secrets.** None.

## Not verified without a device

The atlas paints on a real `<canvas>`, so the one thing argued rather than seen is
**whether a ten-character numeral reads at speed on a tablet**. Everything about it
is computed here — 0.35 em of advance, ~20 css px of type, 39 rpx of cap height on a
reference phone — and it clears the only floor this repository states for a numeral
on a moving object by a wide margin. But 0.35 em is condensed, it is the narrowest
numeral this game has ever drawn, and a child reading `2,533,990,816` off a moving
orb in a bullet field is the test.

If a playtest says it is too tight, the next lever is identified and costed: the orbs
already weave ±5.5 units vertically with independent phase, so giving the four of them
a deterministic ±6.5-unit stagger would put alternate labels on disjoint bands and
**double** the lane — enough to draw ten characters completely undistorted. It is a
change to how a child threads the orb field, which wants a tablet and a child rather
than a patch, so it is named here rather than done here.

Also unseen: the widened cell in the GL layer (the quad is `iSize * uAspect` and only
`uAspect` moved), and the promoted rows' pacing — `long-multiplication` declares a
75 s median and POLARITY's orb hold is `max(4.6, 7.5 − bearers × 0.35)` seconds, so the
top rung is almost certainly unfinishable inside one Bearer. That is the fleet-wide
timing finding #683 recorded as still open, same owner, and it is not made worse here.

### DIFF
--- a/dynawalla/games/polarity/src/contract.ts
+++ b/dynawalla/games/polarity/src/contract.ts
@@ -16,11 +16,52 @@ export type Question = {
   difficulty: number; // 0..1
 };
 
+/**
+ * What POLARITY asks for when it asks for a question.
+ *
+ * Field-for-field a subset of `packs/shared/game-host`'s `DifficultyRequest`,
+ * which is the object the real host reads — so this stays assignable and the
+ * swap stays a one-line import change.
+ *
+ * **On the scale.** The host reads a value below 1 as a 0..1 fraction and a
+ * value from 1 to 10 as a ladder index, which makes exactly one number
+ * ambiguous — `1` — and it resolves it as the BOTTOM of the ladder. POLARITY
+ * reaches a difficulty of exactly 1 after fifteen strata and means the top, so
+ * it speaks the ladder scale on both fields; see `ladderScale` in `seal.ts`.
+ */
+export type Ask = {
+  domain?: string;
+  difficulty?: number;
+  /**
+   * A ceiling on the same scale. The stream never goes above it.
+   *
+   * This is how a pack that cannot DRAW a rung tells the host, instead of
+   * declining item after item from a rung the host will keep serving. Declining
+   * is per-item and the host serves by rung, so without this a level whose every
+   * item is unprintable is not a graceful degradation — it is a Seal Bearer that
+   * asks nothing, forever, and the child at the top of the ladder is the one it
+   * happens to.
+   */
+  maxDifficulty?: number;
+};
+
 export type Host = {
-  next(opts?: { domain?: string; difficulty?: number }): Question;
+  next(opts?: Ask): Question;
   report(r: { questionId: string; correct: boolean; ms: number; answered: string }): void;
   haptic(k: "light" | "medium" | "heavy" | "success" | "failure"): void;
   prefersReducedMotion(): boolean;
+  /**
+   * Throw away questions prefetched at a difficulty the game has since moved
+   * off. Optional and feature-detected, exactly as `trebuchet` and `siege` do
+   * with their own host extensions: the stub host has no pool to flush and an
+   * older host may not have the method.
+   *
+   * A lowered ceiling needs it. The pool holds up to sixty-four questions and
+   * they were all drawn from the rung that has just been ruled out, so without
+   * a flush the ceiling takes sixty-four questions to arrive — which for a
+   * Bearer every twenty-six seconds is most of an afternoon.
+   */
+  flush?(): void;
 };
 
 export function mount(_el: HTMLElement, _host: Host): { unmount(): void } {
--- a/dynawalla/games/polarity/src/core/labels.test.ts
+++ b/dynawalla/games/polarity/src/core/labels.test.ts
@@ -4,12 +4,22 @@ import assert from "node:assert/strict";
 import {
   LABEL_ASPECT,
   LABEL_CAPACITY,
+  LABEL_CAP_RATIO,
+  LABEL_CELL_H,
+  LABEL_COLS,
+  LABEL_ROWS,
+  LABEL_EM,
   LABEL_FAULT,
+  LABEL_INK_W,
+  LABEL_MAX_CHARS,
+  LABEL_MIN_ADVANCE_CAPS,
+  LABEL_MIN_ADVANCE_EM,
   LabelBook,
   isPrintable,
+  labelAdvanceEm,
   labelText,
 } from "./labels.ts";
-import { BULLET, HALF_W } from "../game/constants.ts";
+import { BULLET, HALF_W, MAX_HALF_H, MIN_HALF_H, ORB_SPREAD } from "../game/constants.ts";
 import { orbValues } from "../game/seal.ts";
 import type { Question } from "../contract.ts";
 // The REAL stream. `ladder()`, `choicesFor()` and `answerText()` are the host's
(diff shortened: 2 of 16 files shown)


skyl:
**Auto-merge disabled. Do not merge until the HIGH below is fixed.**

An adversarial review found a HIGH that CI's `adversarial-review` gate did **not** catch (it passed). The authoring agent died on an API 500 before it could act on the review.

## HIGH — `capBelow` pins the whole run to the bottom rung after one dry pool

`src/game/seal.ts:204-211` calls `capBelow(w, q)` on **every** `orbValues` refusal. But `orbValues` returns `null` for four reasons and only two are properties of the rung. The one that matters is the "fewer than two values" branch (`seal.ts:146-161`) — and the real host's dry-pool sentinel trips exactly that:

`packs/shared/game-host/index.ts:681-684` returns `{ id: "", answer: "0", distractors: [] }`. That parses and prints fine, yields one value, so `orbValues` → `null` → `capBelow` reads `q.difficulty === 0` → `drawCeiling = 0`, and the monotone guard at `seal.ts:192` makes it **unraisable**. `startRun` deliberately does not reset it (`sim.ts:755-758`), so it survives every restart for the life of the mount.

Reproduced against the real `game-host` + real `stepBoss`/`drawAskable`:

```
after the dry Bearer, drawCeiling = 0
rung ordinates served after recovery: -1, 0, 0, 0, 0, 0
```

**Effect:** one transient `items.next` → `null` on the first fill, and every question for the rest of the session is the easiest rung in the product. Before this PR the sentinel was merely declined and the run continued at the right difficulty — so this is a regression, and it is the exact opposite of the adaptive behaviour the product is being built around.

Two minimal fixes suggested by the reviewer: don't cap on the "fewer than two values" branch (an item fact, not a rung fact), and skip `capBelow` when `q.id === ""` (the host's own marker for "not a served item").

Also same line: `capBelow` is NaN-transparent — a `NaN` difficulty sets `drawCeiling = NaN`, which `readScale` discards, so the ceiling is inert **and** `w.drawCeiling <= capped` is permanently false, re-logging and re-flushing on every refusal.

## Also flagged (not blocking)

- The `flush()` ordering is backwards: it fires **before** the new `maxDifficulty` is sent, so `flushNow` computes `distance()` against the stale target and *keeps* the banned rung's questions. Measured on the real host: ten consecutive Bearers ask nothing (~4.5 minutes of silence) before it converges. The test's "costs one Bearer, not one per Bearer" claim holds only inside the fake host, which has no pool.
- The `LABEL_MAX_CHARS` derivation is correct for the *launch* geometry, but `stepOrb` decays velocity and the orbs weave — measured over 13,560 settled frames, numeral boxes overlap in 100% of frames at 10 chars. Not a regression (pre-PR is 99.9% at 4 chars), but the central legibility test validates a model the running game does not obey.
- The widened `maxChars` parameter has **no caller anywhere** in `src/`, and both docblocks cite a test that does not pass it.

## Verified correct, do not re-derive

`LABEL_MAX_CHARS = 10` and 11 not fitting; whole-texel atlas under 2048 on both tiers; cap height clearing 22rpx on every viewport; `CEILING_STEP` excluding exactly one rung; the promotion being arithmetically safe (widest integer answer in the entire graph is 10 chars, `1042977861`); PR #660's three guards intact; no security or pack-compat findings. An 8-mutation battery confirms the tests are not vacuous.

skyl:
HIGH fixed in `9b0efc8d0`, along with both MEDIUMs that were mechanical.

- **HIGH** — `rungCannotDraw(q)` now decides whether to cap, and asks only the question that describes a rung (is the ANSWER printable). Too few distractors is an item fact and no longer caps. Verified by deleting the fix: the new test reports the ceiling the dry pool set and lists the bottom rungs served after recovery.
- **NaN** — `capBelow` returns early on a non-finite difficulty. Verified by deletion.
- **Flush ordering** — the flush now waits for `askShape` to carry the new ceiling into `next()`, via `World.pendingFlush`. `flushNow` ranks with `distance()`, which reads the host's own `ceiling`, and that is only updated inside `next()` — so flushing from `capBelow` ranked against the stale ceiling and kept the banned rung.

Not addressed, deliberately, and each is a note rather than a defect in this diff:

- The `LABEL_MAX_CHARS` derivation models the launch geometry while `stepOrb` decays velocity and the orbs weave. The reviewer measured 100% numeral-box overlap at 10 chars in settled frames — **and 99.9% at 4 chars pre-PR**, so it is not a regression and the field already looks like this. Worth its own change; it is a claim about what the test establishes, not a bug this PR introduces.
- The widened `maxChars` parameter still has no caller. Left in place as the seam the refusal tests would need to narrow the budget, but the docblocks' claim that a test uses it is false and should be corrected or the parameter dropped.
- `id === ""` is kept as belt-and-braces and is documented in source as **not** load-bearing today, since the sentinel's answer prints.

62 tests, five consecutive runs, `tsc` clean, on Node 24.18.0 to match CI.

PR #712, opened 2026-07-30
dynawalla/games: say what a keep is, and every other invented word, where a child first meets it

skyl:
> "The instructions need to define the terms sometimes. I don't know what 'the fall' is on the grapple foundry for example. We might need a global pass on the instructions to explain any jargon the game has more plainly."

He is right, and it was systemic. The manuals were already written in short plain sentences — that part was fine — but they spend the flavour words as if the child already owns them. A child who does not know a word cannot follow the sentence it is in.

**Copy only. Every changed line in this diff is a string literal** — no mechanic, layout or constant is touched. 21 of 23 games edited; 126 insertions.

## What changed

**Defined at first use, keeping the flavour word.** A keep is still a keep; it now also says what one is.

| game | terms now defined |
|---|---|
| `trebuchet` | **keep** (was undefined *in the summary*), **the dial** |
| `skyledger` | **astrolabe**, **watch**, **chain**, **MARK**, **MARKS WIDE**, **ONES/TENS**, **logged**, the pre-inked hundreds, and the seven lamps (the manual never mentioned that you can lose) |
| `colossus` | **keystone** (was undefined in the summary), **THE COLOSSUS KNEELS** |
| `coil` | **jaws**, **shear**, **slag**, **brick/course** |
| `street` | **block**, **rivet**, **shutter**, **IN THE STREET** |
| `slice` | **gourd**, **sigil**, **FAVOUR**, **CHAIN**, **lamps** |
| `horde` | **WARDEN**, **CORE**, **CACHE**, **RIFT**, **OVERCHARGE**, **A TIDE**, **RISEN** |
| `polarity` | **vent**, **CLUTCH**, **seal**/**SEALS 4/6**, **STRATUM**, **REPOLARIZE**, **VENT EXACTLY** |
| `beam` | **lattice**, **anchor**, **core**, **RESONANCE**, **SOLE ×2**, **HARMONIC** |
| `forge` | **SEAL**, **PLUNGE**, **quench**, **forge mark** |
| `guilty` | **THE ARBITER** (and that bullets bounce off its shield), **SHUT SHELLS**, **ONE MORE** |
| `runner` | **Voltage**, **Surge**, **OVERDRIVE**, **Recharge**, **gate** |
| `siege` | **ANVIL**, **THE FORGE WENT COLD**, **RELIGHT**, **WAVES HELD**, **DPS**, **HP IN** |
| `pulse` | **REGROUP**, **OVERDRIVE**, **SOLVED/OFF/GONE**, **triplets** |
| `rhythm` | **RESTART THE HEART**, **SPLIT**, **BPM**, **LV**, the sector names |
| `stack` | **True**, **Stratum**, shoring up |
| `arena` | **RESONANCE**, **RANK** |
| `balance` | **counterpoise**, **MOVEMENT IV** |
| `serpent` | **depth**, what a **shield** does |
| `claim` | **VAULT** |
| `merge-idle` | **A SWELL IS RISING**, **ORDERS OF MAGNITUDE** |

`merge` and `mosaic` were already clean and got nothing. `mosaic` draws no English on screen at all and its manual is fully plain; `merge` defines its KEY in its own summary.

## Three copy bugs, not vocabulary

The brief warned about a manual describing something that never existed. Three found:

1. **`stack` promised "Waiting never costs you anything." It does cost you.** `tuning.ts` — `DITHER_CYCLES: 3`, `DITHER_STEP: 0.16`, `DITHER_MAX: 1.9`, with the comment *"After this many full cycles without a drop, the sweep leans on you."* After three idle passes each further pass multiplies the sweep speed, up to 1.9×. The manual now says nothing is taken off the tower and there is no clock — both true — and that the stone starts sliding faster after three passes. Its "the tower falls over and the run is done" was also wrong: **shoring up** has always existed and the manual never mentioned it.
2. **`slice` promised "Thinking is always free, and so is running out of time."** Free of a *lamp*, yes — `lampCost()` returns 0 for all three verdicts, deliberately. Not free of FAVOUR: `favourAfter()` sends **wrong and timeout both all the way back to 1**, equally and on purpose. Said plainly now.
3. **`trebuchet` credited a crew that does not exist** — no entity, sprite or label anywhere in the game. The engine aims into the wind by itself (`sim/verdict.ts` `aimShot`). That is what it now says.

## Not done: the automated gate

The standard asked for a test so game 28 inherits the rule, or a plain explanation. It is the explanation, for two independent reasons — see the review comment below.

## Verification

Node 24.18.0. `tsc` clean and `npm test` green **5× in a row** for all 21 touched games. Untouched: `foundry`, `lattice`, `truedraw`, `counterweight` (agents live), and `packs/**`, `dynawalla-app/**`.

### DIFF
--- a/dynawalla/games/arena/src/mount.ts
+++ b/dynawalla/games/arena/src/mount.ts
@@ -104,7 +104,7 @@ export function mountArena(el: HTMLElement, host: Host, opts?: MountOptions): { 
       {
         heading: "The dark moment",
         lines: [
-          "Every so often the water goes dark and a math question appears above you.",
+          "Every so often the water goes dark and a math question appears above you. The screen calls it a RESONANCE, which is just the name for this moment.",
           "Four glass balls float around you. Each ball holds a different answer.",
           "Swim into the ball with the right answer and you grow a lot.",
           "Pick the wrong ball and you lose some size. The right ball lights up so you can see it.",
@@ -115,6 +115,7 @@ export function mountArena(el: HTMLElement, host: Host, opts?: MountOptions): { 
         lines: [
           "The deeper you go, the more the water changes colour and the more there is to dodge.",
           "The small marks under the depth name fill in as you go down. They never empty.",
+          "RANK counts the numbers bigger than you. RANK 4 means three of them are bigger; RANK 1 means none are.",
           "If something much bigger hits you, you burst and scatter, then come straight back.",
           "There is no ending and no way to lose the run. You play until you want to stop.",
         ],
--- a/dynawalla/games/balance/src/game.ts
+++ b/dynawalla/games/balance/src/game.ts
@@ -177,6 +177,7 @@ export class Game {
             "When the two sides weigh the same, the arm stops tipping and sits flat.",
             "That flat arm is the equals sign. Both sides really do weigh the same.",
             "The words cut into the stone say the same thing with numbers.",
+            "A counterpoise is a weight you put on to balance another weight. That is what you are doing, and it is what the game is called.",
           ],
         },
         {
@@ -200,6 +201,7 @@ export class Game {
           lines: [
             "Nothing buzzes at you when you are wrong. The arm just swings the way you made it swing.",
             "Take the weight off and try a different one. You can try as many times as you want.",
+            "The banner that says MOVEMENT and a Roman number is the stage you have reached. MOVEMENT IV is stage four.",
           ],
         },
       ],
--- a/dynawalla/games/beam/src/mount.ts
+++ b/dynawalla/games/beam/src/mount.ts
@@ -106,7 +106,7 @@ export function mountBeam(el: HTMLElement, host: Host): {
   const guide = createInstructions(root, {
     title: "LATTICE RUNNER",
     summary: [
-      "Five beams of light run down the hall. Each beam has a number at the bottom.",
+      "Five beams of light run down the hall, and together they are the lattice. Each beam has a number at the bottom.",
       "Robots walk down carrying numbers. Ride a beam and shoot — but a shot only works if the beam's number divides the robot's number.",
     ],
     sections: [
@@ -125,6 +125,8 @@ export function mountBeam(el: HTMLElement, host: Host): {
           "A shot destroys a robot only if you can share its number into equal groups of the beam's number, with nothing left over.",
           "Beam 5 takes 405, because 405 is made of fives. Beam 4 does not.",
           "Big robots can often be taken from more than one beam. The biggest beam that fits is worth the most.",
+          "SOLE ×2 means your beam was the only one on the board that fitted, so that robot was worth double.",
+          "HARMONIC means one shot took three robots at once.",
           "A shot that does not work costs you nothing. The robot just rings and keeps walking.",
         ],
       },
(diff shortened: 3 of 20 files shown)


skyl:
## Why there is no test in this PR

The standard offered a door: *"If a robust automated check is not feasible, say so plainly and explain why rather than shipping a test that passes on anything."* I am taking it, on two independent grounds. Twelve vacuous tests in three days is the reason I would rather argue this than add a thirteenth.

### 1. "Defined nearby" is not mechanically decidable, and every regex for it is vacuous

The obvious shape is: term from a jargon list appears in an instruction string ⇒ a definitional connective must appear in that string or the next. The connectives available to a regex are `a X is`, `the X is`, `X are`, `X means`, `called a X`.

That check **passes the exact bad example from the brief**:

> Bad: *"One over the answer and the fall is lost."*

`the fall is` matches `the X is`. So the sentence that motivated this entire task is green under the check. Every weaker pattern is worse and every stronger one needs to know what a category noun is.

### 2. The register problem, in both directions

I tried the register-free formulation too — *every ALL-CAPS word the game literally draws (`fillText` / `textContent` string literals) must appear somewhere in that game's instruction copy*. That is glob-discovered, has no central list, and would genuinely have caught `WARDEN`. It is also unshippable:

- `horde` draws ~30 upgrade-card titles and stat labels — `FEROCITY`, `QUICKENING`, `CARAPACE`, `AVARICE`, `PLATING`, `MOTE`, `VOLLEY`, `CRIT`. Each is self-explaining **on the card it is printed on** (`FEROCITY +25% all damage`). Requiring all thirty in the manual is wrong, and excluding them needs a per-game list.
- `arena` draws 64 decorative rival names (`JETSAM`, `OBOL`, `SPUME`, `UMBRA`). Zero mechanical meaning. Same problem.
- `stack`, `runner`, `pulse` and `rhythm` draw palette/biome/stage/sector names off data tables (`VERDIGRIS`, `THE BLEACH`, `GLACIER`). Flavour.

The inverse check — *every ALL-CAPS word the manual names must be a string the game actually draws*, which is the fabrication class (`stack`'s dither, `foundry`'s lapis slab) — fails on `merge`, whose manual correctly says "the KEY is the big number in the circle" while the circle draws a bare numeral. The manual is *teaching the child the game's word for it*; the check calls that a fabrication.

Either direction needs an exception list. An exception list is a register, and a register is what "adding the thousandth pack is adding a directory" exists to forbid. A curated central lexicon has the same defect one level up: game 28 inherits nothing unless somebody remembers to add its words.

### 3. It also has nowhere legal to live

Separately from feasibility: the check is one module used by every game, so it belongs beside the two surfaces it reasons about, in `dynawalla/packs/shared/game-chrome/` — where CI already runs a globbed shared-module test step. This task is scoped to `dynawalla/games/*/` copy files and explicitly forbids `packs/**`. The alternatives are 23 near-identical per-game test files (the twenty-seven-bespoke-panels antipattern this very module was created to end) or one cross-game test parked inside one arbitrary game's directory, gating the other 22 from a folder that does not own them.

### What I would build instead

A **rewrite-review gate, not a regex**: `codex exec` as an LLM judge over each game's extracted `InstructionsSpec`, asked one binary question per term — *"reading only this text, could a seven-year-old say what this word means?"* — with the game's `pack.json` and HUD strings as context. Glob-discovered over `dynawalla/games/*/`, no register, no exception list, and it is the only formulation that actually answers the question the brief asks. Binary beats a strict scale, per the Qwen language-support pass. That is a separate PR and a separate argument about running a model in CI, so I have not smuggled it into a copy change.

PR #748, opened 2026-07-31
dynawalla: let the child study the answer — a fleet reveal-patience audit, and the two worst offenders fixed

skyl:
"When you fold you should be able to study the answers and then go on, not just have the answers flashed for a second and then go on. This could be a larger thing to audit."

It was literally a second. THE GAVEL's `revealHoldMs` floors at `MIN_REVEAL_MS = 900`, and ten lots of decent play already puts a child at that floor — so the settled room, the only place in the game where every rival's sum is ever completed in front of them, was up for nine hundred milliseconds and then gone.

It generalised. **All 28 games audited**, with the constant and file:line for each. Two fixed here, on the founder's standing instruction to validate a shape on a couple of games before generalising.

## The audit

| game | reveal after a miss? | how long, and driven by what | child can dismiss? | frozen underneath? | verdict |
|---|---|---|---|---|---|
| **stack** MONUMENT | yes — sum completed in accent | `revealDwell(streak)` 4.2 s → 0.9 s, **shared curve** `game/tuning.ts:171` | **yes** `sim.ts:292` | yes, total `sim.ts:319` | **GOOD** (ref) |
| **slice** | yes — plate completes itself | `revealDwellSeconds(intensity)` 4.2 → 0.9 s, **shared curve** `sim/economy.ts:208` | **yes** `mount.ts:384` | mostly `mount.ts:1469` | **GOOD** |
| **arena** | yes — finished sum in ribbon | `world.revealSeconds` 4.2 → 0.9 s, **shared curve** `sim/world.ts:1025` | no | yes `world.ts:2147`,`:1601` | **GOOD** |
| **beam** | yes — sum completes on the wall | `revealSeconds(item)` 1.5–3.0 s, fn of prompt width `sim/window.ts:140` | no — input *refused* `mount.ts:813` | yes, total `mount.ts:875` | **GOOD** |
| **gavel** ⬅ | yes — every rival's value | ~~`revealHoldMs(intensity)` 4.2 s → **900 ms floor**~~ → **waits** | ~~tap shortens only~~ → **yes, and it is the only end** | yes | **FIXED** |
| **truedraw** ⬅ | yes — slate finishes the sum | ~~**900 ms flat**, ×0.45 animation ⇒ **~495 ms** readable `game/round.ts:114`~~ → **waits** | ~~no~~ → **yes** | yes | **FIXED** |
| **horde** | 3 paths, 3 policies | core orb **620 ms** `game.ts:1522`; rift **900 ms** `:1654`; sealed cache **unbounded** `:1598` | sealed yes; core/rift no | core: **9 enemies spawn onto the player during it** `:1512`; rift: death clock runs **and is docked 4 s for the wrong answer** `:1646` | **WEAK** |
| **foundry** | yes — `it needed 12` | 1.25 s `mount.ts:284` | **backwards** — a tap *shortens* it to 0.16 s `bout.ts:333` | count stops, but banner decouples and prints over the *next* prompt | **WEAK** |
| **merge** | yes — `String(r.target)` in red | **0.7 s** `game.ts:1047`, already ~30% faded when it goes `render.ts:926` | no | board yes; but a **9 s answering countdown** precedes it `game.ts:99` | **WEAK** |
| **siege** | yes — `.reveal` outline ×2 | anvil `QUENCH_SECONDS` 1.15 s `constants.ts:173`; focus **`setTimeout(…, 520)`** raw literal `mount.ts:712` | no | **no** — `step()` runs unconditionally `mount.ts:519`; and the revealed slug is dimmed to 45% brightness `styles.css:516` | **WEAK** |
| **runner** | yes — correct value floats up | **1.6 s**, 9th positional arg `game/mount.ts:705` | no — input *disabled* 0.85 s `pacing.ts:51` | **no** — `collide()` runs `mount.ts:1147`; next prompt overwrites the HUD 0.3 s in `:1161` | **WEAK** |
| **mosaic** | Forge beat only; main wall none | `forge.resolving = 1.15` bare literal `game/forge.ts:96` | no | **no** — 12% time-scale `mount.ts:208`; timeout path shows *nothing* `forge.ts:135` | **WEAK** |
| **forge** | partial — correct ingot flares, **no equation** | 0.6 s `forge.ts:1013`; flare decays in 0.42 s, prompt shatters in 0.31 s | no | no clock against them | **WEAK** |
| **trebuchet** | banner — **a visual no-op on waves 1–8** | `revealT = 1.6` `game.ts:1093`; `showBanner` masks it `render/pieces.ts:388` | no | **no** — the ram advances through `settle` `world.ts:456`; next question stamped ~0.5 s before the reveal ends | **WEAK/none** |
| **lattice** | no — `"NOT YET"` | `REFUSE_COOLDOWN_MS = 900` `arena.ts:181` | hint tree is unbounded, but a refusal does not trigger it | yes in effect — no clock, primes come back | **WEAK** |
| **balance** | **no** — weight tossed back | 0.85 s phase `game.ts:531` | n/a | yes — no clock exists `pacing.ts:15` | **MISSING** |
| **merge-idle** | **no** — floater says `"3 BACK"` | none `game.ts:775` | n/a | yes — target explicitly not replaced `engine.ts:359` | **MISSING** |
| **street** | **no** — rivet caves in | `TIMING.rivet = 260 ms` `core/feel.ts:59` | n/a | yes — genuinely no clock `street.ts:295` | **MISSING** |
| **colossus** | **no** — `cursor += 1` in the same call | none `game/game.ts:196` | n/a | no clock | **MISSING** |
| **coil** | **no** — a duller 260 ms "slip" | `REACTIONS.slip.budgetMs` `reactions.ts:28` | n/a | item replaced 0.52 s later | **MISSING** |
| **claim** | **no** — `plates = []`, `hideCard()` same frame | none `game/index.ts:1121` | n/a | n/a | **MISSING** |
| **counterweight** | **no**, and it shows **the wrong number**: `"SHORT — YOU WROTE 42"` | `TIMING.settleMs = 1150` `bout.ts:117` | no | yes | **MISSING** |
| **serpent** | **no** — floater is `−3`, the *cost* | `hitstopWrongMs: 110` `tuning.ts:85` | n/a | **no** — arena closes, and orbs **relabel on a timer** `world.ts:300` | **MISSING** |
| **rhythm** | **no** visual — a ghost bell, `if (this.soundOn)` | ~1 bar `core.ts:664` | n/a | **no** — chart scrolls, countdown bar drawn `renderer.ts:948` | **MISSING** |
| **pulse** | **no** — a red `"OFF"` banner | `gateGlow = 0`, prompt gone in a frame `scene.ts:270` | n/a | **no** — music transport never pauses | **MISSING** |
| **guilty** | **no** — buzz, red flash, husks dissolved | `Phase.Breach` 1 s `game.ts:746` | n/a | **no** — and `world.descent *= 1.14` **escalates on failure** `game.ts:379` | **MISSING** |
| **polarity** | **no** — the *prompt itself* is deleted `renderer.ts:615` | n/a | n/a | **no** — a 14-bullet radial barrage fires on the wrong answer `seal.ts:429`; window shrinks 7.5 s → 4.6 s | **MISSING** |
| **skyledger** | **no** — a cold ring at the station **the child named** | `COLD_LIFE_MS = 420` `scene.ts:52` | n/a | **no** — stars keep falling `game.ts:321` | **MISSING** |

**14 of 28 never show the answer at all. Not one game in the fleet let a child hold it as long as they wanted** — every duration was an auto-dismiss timer, and `beam`, `runner` and `siege` actively *refuse input* during it. Only `stack` had a tap-to-clear path, and even that only *shortened* the beat.

## The shared primitive

`packs/shared/game-pacing` gains `revealPlan(spec, intensity)`:

```ts
export function revealPlan(spec: FlowSpec, intensity: number): RevealPlan {
  if (!revealShown(spec, intensity)) return { settleMs: 0, holdMs: 0 }
  return { settleMs: REVEAL_SETTLE_MS, holdMs: Number.POSITIVE_INFINITY }
}
```

**A shown reveal never expires.** A child who has just been told they are wrong is the slowest reader in the session, and a timer sized for a fluent one removes the evidence exactly when it was about to be useful.

**The adaptation survives, on the same scalar, inverted — it moved from the reveal's LENGTH to its EXISTENCE.** `revealShown` already drew the line at 250 ms, so above intensity ≈0.75 the plan is no reveal at all and an immediate move on; skipping the ceremony is still the reward for mastery. What is gone is the middle — the half-patient reveal long enough to be noticed and too short to be read, which served nobody at either end.

`REVEAL_SETTLE_MS = 350` is the other half, and it is latency rather than pedagogy. A reveal is put up by the same gesture that ended the question, and on a touch screen that gesture is routinely still arriving. Without a floor, the second tap of a double-tap lands inside the reveal's own fade-in.

Both games decide *when* to ask for a plan. A reveal after a clean win would turn a game into a queue of dismissals.

## THE GAVEL

`teaches()` decides which lots wait: everything but a lot bought at a profit with the sum right — the fold, being outbid, paying over the offer, selling with nothing in it, and any wrong assertion. Nothing runs underneath a held room: no lot, no question, no report, no skip, no ladder step, no coins. A brass hairline fades in under the gallery once a tap would actually be honoured — a mark, not a sentence, and not a countdown.

## THE TRUE DRAW

A new `study` phase follows `verdict` on a miss: `verdict` runs the cross-fade that completes the sum, and `study` is that finished frame standing still with no deadline on it, ended by `Round.tap`. The scalar is the streak of correct calls **carried into** the miss — MONUMENT's signal, because tenure is one-way and a streak is not.

Two real bugs found while testing it:
- Reading the *live* streak gave the full patient hold to a child eight true calls into a flawless run, because a miss zeroes it before the beat is chosen.
- `SETTLED_PHASES` had to list `study` in four render branches, or the hold holds a **blank slate showing the wrong claim** — worse than the flash it replaced. The mutation says so in as many words: `the true value is not on the slate while it is being held: 4003 − 87 = 39260`.

## Deferred, deliberately

Not retuning 26 games blind. Ranked by harm — the first four **run a clock against a child who is reading, or escalate because they were wrong**:

1. `guilty` — `world.descent *= 1.14` on a wrong answer. Less time to reason, as a consequence of not knowing.
2. `horde` rift — `riftT -= 4` for the wrong answer, and the death clock runs through the 900 ms reveal.
3. `skyledger`, `polarity`, `serpent`, `rhythm`, `pulse` — no reveal, and nothing stops.
4. `trebuchet` — the ram advances through `settle`, and the reveal is invisible on beginner waves anyway.

Then the cheap ones: `counterweight` and `skyledger` show the child's *wrong* number where the true one belongs; `merge`, `foundry`, `siege`, `runner`, `mosaic`, `forge` all have a working reveal and only need `revealPlan` plus a dismiss path. `truedraw`'s `lapse` still shows nothing — a timeout draws no `flourish`, so holding it would hold a blank correction; that reveal has to be built, and it is asserted as a known gap rather than left silent.

## Verification

- `tsc` and 5 consecutive suite runs on every touched package: game-pacing 23, gavel 84, truedraw 346 — 0 failures each time.
- Downstream consumers of the shared module re-run green: stack 65, slice 137, arena 96.
- Every new assertion mutation-checked. The transcript is in the review comment below.


skyl:
## Mutation transcript

Every new assertion was checked by breaking the line it covers and confirming it FAILS, with the exact message. All restored.

**Shared — `packs/shared/game-pacing/flow.ts`**

| # | mutation | failing assertion |
|---|---|---|
| M6 | `holdMs: Number.POSITIVE_INFINITY` → `holdMs: revealMs(spec, intensity)` (the old finite countdown) | `at intensity 0.00 the reveal expires by itself after 4200ms` · `at 0.72 the reveal locks input for 350ms but only lasts 329.2799999999993ms` · and in gavel: `a fold was settled with no reveal to study`, `the settled room was gone 4200ms after the fold — the old timer is still in there`, `the settled room expired while the manual was up` |
| M7 | `REVEAL_SETTLE_MS = 350` → `0` | `a 0ms floor is inside a double-tap and would be eaten by one` · and in gavel: `an in-flight tap took the reveal down inside its own fade-in`, `half a minute behind a sheet paid off the settle floor` |

**THE GAVEL — `games/gavel/src/game/auction.ts`, `src/render/scene.ts`**

| # | mutation | failing assertion |
|---|---|---|
| M1 | delete `if (this.waiting) return []` from `advance` | `the settled room was taken away while it was being read` · `the settled room was gone 900ms after the fold — the old timer is still in there` · `the room a child at the reveal floor folded on was flashed away` · `the first tap back from a pause skipped the reveal` · `no new lot was ever called` · `the settled room drew no verdict` (7 failures) |
| M2 | delete `if (this.elapsed < this.settleMs) return []` from `nudge` | `an in-flight tap took the reveal down inside its own fade-in` · `the first tap back from a pause skipped the reveal` |
| M3 | `teaches(settled) ?` → `false ?` (no lot ever teaches) | `a fold was settled with no reveal to study` · `the settled room was gone 4200ms after the fold` · `half a minute behind a sheet paid off the settle floor` · `the settled room expired while the manual was up` |
| M4 | `teaches`: exclude `"folded"` | `a fold has nothing to study — the founder's own case` · `a fold was settled with no reveal to study` |
| M5 | `teaches` → `return true` (every lot teaches) | `a clean profitable sale is being held up as a lesson` · `a clean sale held the child for a receipt` |
| M8 | `Scene.onward` returns immediately | `a room with no deadline on it never said it was waiting` |
| M9 | cue ignores `view.nudgeable` | `the go-on mark appeared inside the settle floor, where a tap does nothing` |
| M10 | cue ramps unconditionally | `the go-on mark was drawn over a live room the child is still bidding on` |

**THE TRUE DRAW — `games/truedraw/src/game/round.ts`, `src/render/scene.ts`**

| # | mutation | failing assertion |
|---|---|---|
| T1 | `if (this.holdsAfterVerdict())` → `if (false)` | `a miss went to clear instead of holding` · `the slate was gone 3600ms after the miss — the old timer is still in there` · `an in-flight touch took the lesson down inside its own beat` |
| T2 | delete `if (this.elapsed < REVEAL_SETTLE_MS) return []` from `tap` | `an in-flight touch took the lesson down inside its own beat` |
| T3 | `studyIntensity(this.carriedStreak)` → `studyIntensity(this.correctStreak)` | `a child eight true calls into a run was held for a reveal` |
| T4 | `holdsAfterVerdict` drops the `!isMiss(...)` guard | `a correct call held the child for a receipt` · `the street never reached call in 8s — it is stuck in study` |
| T5 | `drawOnward` returns immediately | `a held sum past its settle floor drew 0 go-on marks against 0 inside it` |
| T6 | `drawOnward` ignores `REVEAL_SETTLE_MS` | `a held sum past its settle floor drew 1 go-on marks against 1 inside it` |
| T7 | `SETTLED_PHASES` (scene) drops `"study"` | `the true value is not on the slate while it is being held: 4003 − 87 = 39260` |
| T8 | `SHOWING` (round) drops `"study"` | `the slate stopped showing the miss` |

**Two things this process caught that a reading would not have**

- **M1 originally passed with the guard deleted.** `waiting` and an infinite `duration` were two mechanisms enforcing one rule, so neither could be tested and one of them would have quietly rotted. Collapsed to one, and M1 then killed with 7 failures. The same audit removed `RevealPlan.shown`, which THE GAVEL consumed *instead of* `holdMs` — so the whole never-expires claim could have been deleted from the shared module without a game noticing.
- **T4 first "passed" by hanging for two minutes.** A change that holds a beat forever makes an unbounded `while (round.phase !== "call")` immortal, and a test that never returns reports nothing. The helper is bounded now and says `the street never reached call in 8s — it is stuck in study`.

No assertion in this PR uses `assert.deepEqual(x, [])`.

PR #765, opened 2026-08-01
dynawalla/games/mosaic: MOSAIC — the window is still being built while you are breaking it

skyl:
> "mosaic is pretty nice but I think we need more juicy remixing of the targets while it's running.. The first board is just a rectangle of the numbers and it's boring. Maybe if it was more sparse or things dropped in or more random .. but it just calmly knocks out all of the even numbers and cracks all of the odd ones and just bounces around, not that fun."

He was describing a fact, not a feeling. Measured over 400 seeds before this PR:

| | before | after |
|---|---|---|
| opening board: distinct shapes / 400 seeds | **1** | **386** |
| opening board: fill fraction | **1.000** (always) | **0.600** avg, 0.44–0.92 |
| opening board: layout families | 1 (`solid`, hard-coded) | 4 |
| wave 5: distinct shapes / 400 seeds | 14 | 399 |
| wave 5: fill fraction | 0.794 | 0.554 |
| changes to a board mid-play | **0, ever** | **15.4 beats/wave, one every 6.2 s** |
| waves with zero mid-play change (of 240) | 240 | **0** |
| panes dropping in per wave | 0 | 11.1 |

## What I built

**THE CARVE** (`wall.ts`). The nine hand-written layout masks now only say what *kind* of window this is; a seeded carve says which one — one to three elliptical voids plus a scatter of single missing panes. Every cut is made at a cell **and at its mirror**, so the result is still a designed rose window rather than a chewed edge, and the two existing mirror-symmetry tests still pass unchanged. A cut that would take the wall below a floor of 20 panes is rolled back whole. The opening wave also draws from a family of shapes that stay legible at four or five rows instead of being hard-coded to `solid`.

**THE REMIX** (`remix.ts`, new). A seeded, jittered scheduler fires a beat every 4.5–7.5 s of wave time:

- **RE-GLAZE** — a mirrored pair of panes falls back into empty cells and sets there. The window fills in behind you, so a half-emptied board is a *different* board rather than a smaller one.
- **KINDLE** — a mirrored pair of *stone* catches light and becomes a target. The field of targets is redrawn under a wall you thought you had read.
- **TURN** — the whole rose takes a new amplitude and period of sideways drift. The maths does not change; the aim does.

A beat with nothing to do falls through to the next kind rather than being spent on nothing.

## Not breaking the maths

Three rules, each a property test over many seeds with **every invariant checked on every frame** of a played-to-completion run (`play()` in `remix.test.ts`), including a six-wave session so the late waves — descent, hardest drift, crystal and star — are covered too.

1. **Nothing a child is aiming at is ever taken away.** A live target never stops being a target, never stops being alive, and never changes its face or its cell. Kindling is strictly stone → glass; there is no path back. Re-glazing only ever writes into a cell that is *empty*. Every remix is additive from where the child is standing — the worst case is a gift, never a tile that vanished from under a shot. This is the ABYSSAL BLOOM / MATH NINJA class and it is the first thing the test asserts.
2. **The board provably converges.** Targets added over a wave are hard-capped at ~⅓ of the opening count, and none may be added once three or fewer are left standing. Total targets a wave can ever contain is `guiltyTotal + budget`, a finite number, and every break removes one permanently. A wave cannot be remixed into being unwinnable *or* endless. A falling pane counts as alive for the clear check but is not collidable, so a wave cannot declare itself clear while glass is still in the air.
3. **It is not a clock.** Every beat is wall-side. Nothing shortens the time a child has to think about anything; the forge is untimed apart from its own constant, generous, unpunished close. Drift is bounded *per wave* — a fifteenth of a cell on the tutorial, a third of a cell late — and the global bound is chosen so a full swing still fits inside the wall margin. Speed stays rewarded (the clear bonus) and nowhere enforced.

Stone re-glazing gets its own budget for pacing: uncapped it took the mean headless wave from 75 s to 104 s while adding no maths at all. Capped, it costs 20 s and the window still visibly fills back in eleven times a wave.

## Also: the miss now holds

The forge is the one place MOSAIC asks a question, and a wrong answer used to take the lit correct rune away after 1.15 s. It now uses `packs/shared/game-pacing`'s `revealPlan` contract: `holdMs: Infinity`, ended by the child's own hand anywhere on screen, with the shared 350 ms settle lockout so the second tap of an impatient double-tap cannot eat the lesson it just raised. The prompt and the correct rune stay up together — the sum completed in front of them, in the warm accent, never in red. A clean win still just plays its flourish and goes; there is nothing to marinate on in a sum you got right. Dimmed runes now fade *down and stay down* instead of quietly returning to full a second later.

## What I rejected

- **Changing the rule mid-board.** The loudest possible remix and the worst one: it re-classifies every tile at once, so a child mid-shot is aiming at something that is no longer a target. Rule 1 forbids it.
- **Guilty → stone (un-kindling).** The obvious symmetric partner of KINDLE, and exactly the bug that cost two rounds on ABYSSAL BLOOM. Kindling is one-directional by construction, and the mutation that widens its predicate to include live targets is killed by the test.
- **Moving individual tiles / sliding rows.** Genuinely juicy, but it makes a tile the child is tracking arrive somewhere else, and it wrecks the mirror symmetry that is the art direction. Drift of the *whole* window gets the "the field moves" feeling from one number, changes no tile's identity, and cannot desync the grid.
- **Tuning the target budget down to shorten waves.** Measured: 0.25 / 0.30 / 0.35 give 105 / 109 / 104 s. The added targets are free; the length came entirely from re-glazed *stone*, so that is what got a budget instead.
- **A new audio cue.** MOSAIC has a private synth predating the shared soundscape and migrating it is not this PR. The two new beats reuse the existing `charge(fraction)` pip — an existing cue, and one where the caller passes a fraction rather than naming a pitch.
- **Sway on the tutorial at full strength.** Wave one gets a fifteenth of a cell: visibly alive, negligible to aim at. A test asserts the tutorial swings less than a tenth of a cell and a late wave more than three times as far.

## Verification

- `npm run -s tsc` clean; `npm run -s test` **98 pass / 0 fail, five consecutive runs**.
- `node dynawalla/packs/build.mjs mosaic` builds and passes the SDK checker.
- Solvability property-tested: 24 seeds played headless to a clear with every invariant checked every frame, plus 4 seeds × 6 consecutive waves.
- **Every new assertion mutation-tested.** 27 mutations applied to production code; all killed but one, which is provably equivalent (`t.guilty = true` vs `t.guilty = !t.guilty` inside a loop whose selection predicate already guarantees `!t.guilty` — the mutation that *widens* that predicate is killed). Transcript in the task report.

### DIFF
--- a/dynawalla/games/mosaic/src/dev.ts
+++ b/dynawalla/games/mosaic/src/dev.ts
@@ -13,8 +13,9 @@ import { createStubHost } from "./stubHost.ts";
 import type { Sim } from "./game/state.ts";
 import { VW } from "./game/state.ts";
 import { buildWave } from "./game/wall.ts";
-import { chooseShard } from "./game/forge.ts";
-import { launch, paddleHalf } from "./game/sim.ts";
+import { chooseShard, dismissForge } from "./game/forge.ts";
+import { createRemix } from "./game/remix.ts";
+import { launch, paddleHalf, wallLeft } from "./game/sim.ts";
 import type { Host } from "./contract.ts";
 
 const el = document.getElementById("stage");
@@ -43,6 +44,12 @@ if (dev && q.get("wave")) {
   const wave = buildWave({ seed: dev.sim.seed, index });
   dev.sim.wave = wave;
   dev.sim.rule = wave.rule;
+  // The remix is per-wave state. Jumping the wall without rebuilding it left
+  // wave 21 running wave 1's ceiling, which is below its own pane count, so
+  // re-glazing bailed on its first line every time and the wave you jumped to
+  // was the one wave in the game that never remixed.
+  dev.sim.remix = createRemix(dev.sim.seed, wave);
+  dev.sim.sway = 0;
   dev.sim.broken = 0;
   const grid = new Int32Array(wave.cols * wave.rows).fill(-1);
   dev.sim.cellW = (VW - 100) / wave.cols;
@@ -62,6 +69,13 @@ if (dev && q.get("bot")) {
       return;
     }
     if (sim.forge) {
+      // A miss holds its reveal for ever. The bot reads it for a second and
+      // taps, the way a child does; without this it sits on the first wrong
+      // answer of the run and the watch loop is dead.
+      if (sim.forge.held) {
+        if (sim.forge.age > 1.2) dismissForge(sim);
+        return;
+      }
       // Answer correctly nine times in ten, so both outcomes get exercised.
       if (sim.forge.age > 1.2 && sim.forge.resolving <= 0) {
         const i = Math.random() < 0.9 ? sim.forge.shards.findIndex((s) => s.correct) : 0;
@@ -84,8 +98,9 @@ if (dev && q.get("bot")) {
     let best: { x: number; y: number } | null = null;
     let bestD = Infinity;
     for (const t of sim.wave.tiles) {
-      if (!t.alive || !t.guilty) continue;
-      const tx = sim.wallX + (t.col + 0.5) * sim.cellW;
+      // `drop > 0` is a pane still in the air: not solid yet, so not aimable.
+      if (!t.alive || !t.guilty || t.drop > 0) continue;
+      const tx = wallLeft(sim) + (t.col + 0.5) * sim.cellW;
       const ty = sim.wallY + sim.descent + (t.row + 0.5) * sim.cellH;
       const d = Math.hypot(tx - landing, ty - sim.paddleY);
       if (d < bestD) {
--- a/dynawalla/games/mosaic/src/fx/render.ts
+++ b/dynawalla/games/mosaic/src/fx/render.ts
@@ -16,7 +16,8 @@
  */
 import type { Sim } from "../game/state.ts";
 import { VW } from "../game/state.ts";
-import { MOLTEN_AT, paddleHalf, tileX, tileY, TRAIL_LEN } from "../game/sim.ts";
+import { MOLTEN_AT, paddleHalf, tileX, tileY, TRAIL_LEN, wallLeft } from "../game/sim.ts";
+import { DROP_SECONDS } from "../game/remix.ts";
 import { MASONRY_HP } from "../game/wall.ts";
 import { ruleBanner } from "../game/rules.ts";
 import { FORGE_TIMEOUT } from "../game/forge.ts";
@@ -461,7 +462,7 @@ export class Renderer {
   /** The stone frame the window sits in. */
   private drawTracery(sim: Sim): void {
     const g = this.ctx;
-    const x = sim.wallX - 16;
+    const x = wallLeft(sim) - 16;
     const y = sim.wallY + sim.descent - 16;
     const w = sim.wave.cols * sim.cellW + 32;
     const h = sim.wave.rows * sim.cellH + 32;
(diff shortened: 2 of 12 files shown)


skyl:
## Adversarial self-review found two real regressions in my own reveal change — fixed in `04b981dc8`

I ran the diff through the adversarial reviewer before asking for a look. It found that the *carve* and the *remix* were clean, and that the thing I broke was the bit I added as a guard-rail.

**HIGH — a held reveal ran the game underneath itself.** The forge has always stepped the sim at 0.12× while a beat is open, which is harmless when a beat lasts at most 7 s. Unbounded — with the manual now telling the child "look at it for as long as you like" — the ball keeps travelling and the wall keeps creeping with their hand off the glass. Measured, 30 seeds × 4 paddle positions:

| held for | runs that lose a bead |
|---|---|
| 1.15 s (the old behaviour) | 0 / 120 |
| 3 s | 24 / 120 |
| 10 s | 81 / 120 |

Median first loss 7.4 s. On the last bead the run *ended* under the reveal, and the dismissing tap then hit the game-over branch and restarted the run instead of taking the reveal down. `step()` now returns immediately while a reveal is held — nothing moves at all — and `act()` gives a held reveal the tap before game-over can claim it.

**HIGH — the answer drifted off the bottom of the screen.** The runes rise on an arc with gravity under them and `stepForge` integrated them *before* the held guard. At `vh = 1000` the lit correct rune left the screen 21 s into a hold advertised as unlimited. They now freeze exactly where they were chosen; only `age` keeps running, because the settle floor is measured in it.

Both are mutation-tested: removing either guard fails with `the ball travelled under a held reveal` / `the correct rune drifted while it was being read`.

**Also fixed:** `MAX_SWAY_CELLS` 0.42 → **0.26** — the bound was checked against the tiles, but `drawTracery` draws 24 units of stone frame outside the grid, so a full swing clipped the frame 8 units off the left edge. The bleed is now a named constant shared by the renderer and the bound, and the test asserts the *frame* fits. `?wave=` and `tools/bench-sim.mts` swapped the wall without rebuilding the per-wave remix, so a jumped-to wave ran wave one's pane ceiling — below its own pane count, which meant re-glazing bailed on its first line and the benchmark could not measure the thing it is cited for. The `?bot=1` autoplayer deadlocked on its first wrong answer (it waited on `resolving <= 0`, now `Infinity`). Both aim helpers used `sim.wallX` instead of `wallLeft(sim)` and treated a pane still in the air as aimable. Two tautological assertions removed (`carved >= 0`, `addedPeak >= 0` — both `Math.max` over monotonic counters).

## Final numbers

`tsc` clean; **100 tests, 0 fail, five consecutive runs**; `node dynawalla/packs/build.mjs mosaic` builds and passes the SDK checker; `npm run -s bench` 0/45 uncleared, median clear 58 s.

| | before | after |
|---|---|---|
| opening: distinct shapes / 400 seeds | 1 | 386 |
| opening: fill fraction | 1.000 always | 0.600 avg (0.44–0.92) |
| wave 5: distinct shapes / 400 seeds | 14 | 399 |
| mid-play changes to a board | 0, ever | one every **6.2 s**, 15.0 a wave |
| waves with no change at all (of 240) | 240 | **0** |
| panes dropping in per wave | 0 | 11.0 |
| targets the remix adds | 0 | 5.2/wave (33% of the opening count, hard-capped) |
| mean headless wave | 74.8 s | 93.5 s |

33 mutations applied to production code, 31 killed. The two survivors are both accounted for: `t.guilty = true` → `t.guilty = !t.guilty` inside a loop whose selection predicate already guarantees `!t.guilty` (equivalent — the mutation that *widens* that predicate is killed), and overshooting the sway target past its own maximum, which the per-frame clamp absorbs (removing the clamp *is* killed).

PR #775, opened 2026-08-03
dynawalla/games/trebuchet: the wind was smaller than the blast, so it hit the right keep and marked you wrong

skyl:
> "for trebuchet, we can't figure out what the wind .. are we supposed to shoot it longer, shorter, ignore it? ... I think in the current implementation they are confusing and don't necessarily do anything."

**Option (a): the wind stays and becomes real maths.** The measurement below is why, and what it cost.

## First, what the wind actually did

Measured on `origin/main` before touching anything — 450 shots across five seeds and waves 1–20, driven through the real `TrebuchetGame` against the faithful ladder harness. The child modelled is a competent one: she reads the sum, works it out **correctly**, and dials her answer without adjusting for the wind.

| wind cap | n | displacement | graze | solid | **miss** |
|---|---|---|---|---|---|
| 3 | 95 | 1..3 m | 72% | 28% | **0%** |
| 4 | 95 | 1..4 m | 55% | 25% | **20%** |
| 0 | 260 | 0 m | — | — | (direct 100%) |

**165 of 190 windy shots — 86.8% — came down inside the blast radius of the keep she was aiming at.** The boulder struck the tower, the tower cracked and leaned, masonry came off it (`game.ts` `struck.damage += 0.34`, `struck.lean`), and the game recorded a wrong answer.

So: is the wind honest? The *arithmetic* was, and #654 fixed that — `claim = dial + wind`, rolled before she commits, never touched again, and this PR does not move any of it. I verified the pacing-audit defect (`landing = R + wind` with the caret on the dial) is genuinely gone.

But the wind's **entire range was smaller than the blast it moved the boulder by** — 1–3 m against `GRAZE_M = 3`, on a field where keeps are 8 m apart and 4.6 m wide. It was invisible in the world and audible only in the verdict. That is strictly worse than either honest option: it looks ignorable and it is not, and the child's only evidence is a tower she watched her boulder hit, still standing, marked wrong.

**Is it learnable?** The compensation is `answer − wind`, one integer subtraction — yes, computable. The reason a child could not learn it was not the arithmetic. It was that there was no observation from which to infer the rule.

### And a second thing nobody had measured

The wind cap by wave, same run:

```
wave  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20
cap   0  0  0  3  3  4  4  0  0  0  0  0  0  3  3  4  4  0  0  0
```

On at wave 4, **off at 8**, back at 14, gone at 18. The wind was gated on the item's difficulty — defensible reasoning — but `stock()` sweeps and *wraps* looking for a rung whose answers fit on 122 metres, so the rung served oscillates and the wind oscillated with it. None of that is anything the child did. A mechanic that appears and disappears is not a difficulty step.

## What changed

**1. The magnitude is pinned to the geometry at both ends** (`sim/ballistics.ts`).

- `WIND_MIN = GRAZE_M + 1` = 4 — one metre past the blast, so an ignored wind lands **clear of her keep** and leaves a crater with her own number in it. That is the feedback the manual has promised for a wrong answer since day one ("you get to see how far off you were") and which the wind was measurably never able to trigger.
- `WIND_MAX = MIN_GAP − 2` = 6 — one metre inside the garrison's reach, so an ignored wind never fires the wrong-horn/counter-volley, which would say "you named the wrong number" when she had named the right one and let the wind have it.

Not claimed: that an ignored wind touches *nothing*. At the closest spacing the field allows, a 6 m wind lands 2 m from the neighbouring keep and the blast shakes it — the truth about the shot, told next to a keep that is not hers. Making it touch nothing needs `MIN_GAP = 10`, and the suite proves the cost directly: mutating `MIN_GAP` to 10 fails with **`wave 3 could not fill the rack`**. The mechanic does not get to shorten the waves.

**2. It is bought, not handed over** (`sim/felled.ts`, `WIND_STEPS` in `sim/world.ts`). A persisted count of felled keeps indexes a monotone three-step ramp — nothing below 12, ±4..5 to 24, ±4..6 above — with the top step being the full mechanic. The first twelve are *necessarily* felled in still air (there is no wind below twelve), so step 1 is bought by fluency at the one-step game and step 2 by fluency at the two-step game.

One counter, not two, and that came from measurement: a first draft counted only still-air kills, and a fresh child playing perfectly **froze at fifteen and could never reach the top step**, because the currency stopped being earnable the moment she crossed the first threshold.

## After

Same harness, a fresh child playing well:

```
waves 1-5   cap 0   still air — 15 keeps felled
waves 6-8   cap 5   ±4, ±5
wave 9+     cap 6   ±4, ±5, ±6      (monotone; never turns off again)

ignored winds that touched her own keep:  0 of 75
garrison misfires:                        0
displacement: 4..6 m, mean 4.9
```

### The compensation a child performs

The boulder says `47 + 25`. The chip shows an arrow pointing downrange with a **5** beside it.

1. `47 + 25 = 72` — the keep she wants.
2. The wind will carry the boulder 5 metres further than the dial says, so take it off: `72 − 5 = 67`.
3. Dial 67, fire. The boulder leaves at 67 and the wind puts it on **72**. The keep comes down.

Forget the 5 and it lands on **77** — five metres past a keep that stays standing, with `77` burnt into the open ground where she can read it. Wrong for a stated reason, not by an invisible margin.

## The angle

Not coming back, and I want to be plain about it rather than quietly ignore half the request. The loft lever was already removed for **exactly the defect being fixed here**: measured over 1,616 wall situations, three of its five settings were identical to the default and the two below it were blocked by the wall 42.9% and 21.9% of the time — every position was either the default or a way to lose. Making a launch angle *computable* by a child in grades 1–6 needs trigonometry that is not in the curriculum, and making it the wind's multiplier stacks a multiplication on top of the sum and the adjustment, three steps deep, on a child who is here to practise `47 + 25`. A second decorative dial is the complaint, not the answer.

## Verification

- `npm run -s tsc` clean; `npm run -s test` **139/139, five consecutive runs**, no flake.
- `node packs/build.mjs trebuchet` builds and passes the SDK checker.
- Adversarial self-review run before opening; all nine findings fixed (manual copy that was false ~10% of the time, stale `dialRange` docs, a `localStorage`/module-state leak across tests, a duplicated `MIN_GAP` in `stubHost.ts`, a junk-value sweep that only tested a weak predicate, a sweep guard sitting on its own boundary).

### Every new assertion mutation-tested

Each mutation applied, suite run, failure and exact message recorded, mutation reverted.

| # | mutation | fails with |
|---|---|---|
| 1 | `WIND_MIN = GRAZE_M` | `landed 17, inside the blast of her own keep` / `the ramp offers 0, 2, 4 magnitudes` |
| 2 | `WIND_MAX = MIN_GAP - 1` | `landed 13, the garrison at 12 fires` |
| 3 | ramp gate opens at 0 felled | `wave 1 put a wind on a child who has felled nothing` |
| 4 | top step goes backwards | `the wind shrank at 24 keeps` |
| 5 | `noteFelled()` never called | `the wind never arrived across 60 shots: 0,0,0,…` |
| 6 | every shot counts, not only kills | `a miss bought progress` / `the wind blew after only 7 keeps` |
| 7 | count never written to storage | `the count never reached storage` |
| 8 | count never read back | `a fluent child must meet a wind; cap is 0` |
| 9 | junk in the slot trusted | `"Infinity" became Infinity keeps` |
| 10 | `rollWind` may emit below `WIND_MIN` | `cap 5 rolled 3` |
| 11 | `windValues` drops the floor | `the ramp offers 0, 5, 6 magnitudes` |
| 12 | `localStorage` access unguarded | `Error: SecurityError: the document is sandboxed` |
| 13 | no `isFinite` guard on the count | `"Infinity" became Infinity keeps` |
| 14 | `installDom` stops resetting the count | `a child on her first wave was handed a wind` |
| 15 | `MIN_GAP` → 10 | `wave 3 could not fill the rack` |

No `assert.deepEqual(x, [])`, and no assertion compares a value to the tuning constant that produced it — the two bounds are asserted through behaviour (`resolve()` over real layouts), which is what mutations 1 and 2 demonstrate.

## Binding

- No clock added, none removed. Speed is not enforced anywhere in this game and this change does not introduce it.
- `windCapFor` is a **pure function of the felled count**, monotone non-decreasing, with the top step being the full mechanic — the ramp shape used in `games/lattice` (#762) and `games/skyledger` (#769).
- Stayed in lane: only the wind and the angle. `src/index.ts` instruction copy is touched only where this change made it false.

Scope is `dynawalla/games/trebuchet` only.

### DIFF
--- a/dynawalla/games/trebuchet/src/game.test.ts
+++ b/dynawalla/games/trebuchet/src/game.test.ts
@@ -1092,6 +1092,15 @@ test('a miss in a wind stops the siege, and hands the wind back when she is read
   assert.ok(runUntil(game, () => reports.length > 1), 'the next shot never resolved')
   assert.equal(reports[1].correct, true, 'the shot after a miss was scored wrong')
   assert.equal(cells.get('dw.trebuchet.felled'), String(FLUENT + 1), 'the keep she felled bought nothing')
+
+  // And across the wave boundary, which is the only place the loss would show.
+  // `windCap` is read once per wave, so a miss that quietly spent the count would
+  // leave this wave blowing exactly as before and take the wind away at the NEXT
+  // one — a mechanic disappearing a minute after the thing that cost her it.
+  assert.ok(dismissReveal(game, canvas), 'a reveal was left standing')
+  game.jumpToWave(4)
+  assert.ok(runUntil(game, () => game.currentPhase === 'aim', 3000), 'the next wave never stocked')
+  assert.equal(game.currentWindCap(), WIND_MAX, 'the wind was gone by the next wave')
 })
 
 test('the wind arrives because she FELLED KEEPS, and only then', () => {
--- a/dynawalla/games/trebuchet/src/game.test.ts
+++ b/dynawalla/games/trebuchet/src/game.test.ts
@@ -16,6 +16,43 @@ import test from 'node:test'
 import type { Host, Question } from './contract.ts'
 import { TrebuchetGame } from './game.ts'
 import { hudLayout } from './render/hud.ts'
+import { resetFelledForTest } from './sim/felled.ts'
+import { WIND_MAX, WIND_MIN, WIND_STEPS } from './sim/world.ts'
+
+/**
+ * The felled keeps this device remembers, set to whatever the test needs.
+ *
+ * The wind is bought with keeps felled and the count is persisted, so a test that
+ * wants the windy game has to say what this child has already done — and a test
+ * that wants the beginner's game has to say she has done nothing. Both go through
+ * the REAL storage the game reads, so the read side is exercised rather than
+ * reached round.
+ */
+function seedFelled(n: number): Map<string, string> {
+  const cells = new Map<string, string>()
+  const g = globalThis as unknown as Record<string, unknown>
+  g.localStorage = {
+    getItem: (k: string) => cells.get(k) ?? null,
+    setItem: (k: string, v: string) => {
+      cells.set(k, String(v))
+    },
+    removeItem: (k: string) => {
+      cells.delete(k)
+    },
+    clear: () => cells.clear(),
+    key: () => null,
+    get length() {
+      return cells.size
+    },
+  }
+  if (n > 0) cells.set('dw.trebuchet.felled', String(n))
+  resetFelledForTest()
+  return cells
+}
+
+/** The first step of the wind ramp, and the top of it, read off the ramp itself. */
+const FIRST_WIND = WIND_STEPS[1].felled
+const FLUENT = WIND_STEPS[WIND_STEPS.length - 1].felled
 
 /* ---------------------------------------------------------------- the glass */
 
@@ -69,8 +106,17 @@ function stubElement(tag: string, w: number, h: number): Record<string, unknown>
 
 const VIEW = { w: 1024, h: 768 }
 
-/** Installs a canvas-shaped world on `globalThis`. Returns the mount element. */
+/**
+ * Installs a canvas-shaped world on `globalThis`. Returns the mount element.
+ *
+ * It resets the felled-keep count too, and it has to: the count lives in module
+ * scope behind a `localStorage` slot, so a test that fells twelve keeps leaves
+ * every test after it in this file starting on the windy game. Eleven tests build
+ * a `TrebuchetGame` directly rather than through `newGame`, and none of them are
+ * about the wind — they would have been silently handed one.
+ */
 function installDom(): { el: HTMLElement; canvas: Record<string, unknown> } {
+  seedFelled(0)
   const created: Array<Record<string, unknown>> = []
   const g = globalThis as unknown as Record<string, unknown>
   g.document = {
@@ -108,10 +154,9 @@ type Reported = { questionId: string; correct: boolean; ms: number; answered: st
 /**
  * A host that serves the answers it is given, on the rung it is told to.
  *
- * The rung matters now: `windCapFor` reads the item's own difficulty to decide
- * whether the wind blows, so a test that wants the beginner's game asks for an easy
- * rung and a test that wants the wind asks for a harder one. Nothing here is a wave
- * number.
+ * The rung no longer decides whether the wind blows — `seedFelled` does, because
+ * the wind is bought with right answers rather than handed over when the curriculum
+ * ladder drifts. The difficulty here is what the questions are, and nothing else.
  */
 function recordingHost(answers: number[], difficulty = 0.4): { host: Host; reports: Reported[] } {
   const reports: Reported[] = []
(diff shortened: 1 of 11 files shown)


skyl:
## Rebased onto current `main` and fixed the merge-queue failure

Green on the branch, red in the queue, because **#774 landed while this was in review**. Rebased onto `origin/main` (now including #771, #773, #774, #776) and reproduced the failure locally on the merged tree — one test, exactly the one reported:

```
✖ only a felled keep buys the wind — a miss buys nothing, and buys it back too
  AssertionError: the next boulder never loaded
```

### What the interaction turned out to be

#774 made a wrong answer **stop the siege**: `revealPlan` supplies `holdMs: Infinity`, and one gate (`stopped`) freezes the phase clock, the dial repeat, the phase machine, the counter-fire and the ram until the child takes the completed sum down by hand. My test drove a miss and then waited for the next boulder. After #774 the boulder correctly never loads — the assertion encoded the pre-#774 world.

The two changes **compose correctly in the sim**; only the test was wrong. `raiseReveal` lives in the `else if (v.report)` branch and `noteFelled()` in the `correct` branch, so they are mutually exclusive — but that was true by reading, not by test, so it is now asserted.

Measured in the merged sim over 69 shots, a child who forgets the wind on every fourth windy boulder:

```
kills 56, misses 13, one boulder swallowed by the ram (correctly unreported)
reveals held after a miss              13 / 13
reveals raised on a felled keep         0
count double-counted / eaten / spent    0
wind changed under a held reveal        0
unbounded waits hit                     0
final felled count 56 = kills exactly, wind cap 6
```

### What changed

- **The failing test now dismisses the reveal the way a child does** — a pointerdown on the open field, past the settle lockout — and asserts *both* sides, so it is stronger than before rather than weakened: the boulder must **not** load while the sum is up (bounded wait, asserted `false`), and must load once it is down. It still proves a felled keep buys wind, a miss buys none, and a windy kill counts toward the top step.
- **New test: `a miss in a wind stops the siege, and hands the wind back when she is ready`.** The wind may not re-roll under a held reveal — a chip that re-rolled would be showing the arithmetic for a shot she has not taken beside the sum for one she has — the ramp may not be spent by being wrong, and the very next boulder must still be winnable.
- **Audited every drive loop in `game.test.ts` for the same assumption** and bounded the two that could exit early in silence. The dial sweep now counts the compensations it actually entered and asserts a floor (`taken >= 8`); the wind-arrival loop counts keeps off the host's **reports** rather than off "a shot went out" — a wrong answer reports too, and would otherwise have been counted as a keep on the ground. Both also assert `revealedSum() === null` at every step, which is what proves every shot in them really was correct.
- Refreshed one comment #774 left pointing at `WIND_FROM_D`, a constant this branch removes.

### Verification

`tsc` clean; **174/174 tests, five consecutive runs**, 5.0 s each (no hang — the suite is bounded and every bound is asserted). `node packs/build.mjs trebuchet` builds and passes the SDK checker.

**Seven more mutations**, each applied, run, message recorded, reverted:

| # | mutation | fails with |
|---|---|---|
| N1 | remove the `stopped` gate | `the next boulder loaded underneath the sum she was reading` |
| N2 | a felled keep also raises the reveal | `a felled keep raised a reveal` |
| N3 | dismissing the reveal counts a keep again | `a felled keep bought nothing` / `a miss spent what she had bought` |
| N4 | the wind re-rolls under the reveal | `frame 0: the wind re-rolled under the reveal` |
| N5 | a wrong answer spends the count back down | `a miss bought progress` / `the reveal moved the count` |
| N6 | the dial sweep exits early | `only 3 compensations were entered` |
| N7 | the arrival loop counts shots, not keeps | `only 3 keeps fell across 3 shots` |

That is 22 mutations across the branch, all biting.

Issue #737, opened 2026-07-30
Soundscape: a generative, in-tune, never-stale audio layer for the whole fleet


skyl:
Design + a working prototype in one game. Full document:
[`dynawalla/docs/SOUNDSCAPE_DESIGN_2026-07.md`](../blob/soundscape/dynawalla/docs/SOUNDSCAPE_DESIGN_2026-07.md). PR: #TBD.

## The founder's idea

> *"we use some stochasticity and randomness .. but in a lot of cases we have a nice drone that is in tune with the little sound effects that play a melody. so for the steelyard, right now have the same sound for every +1/−1 … it would be way cooler if it randomly played a melody based on the randomly chosen soundscape for any given moment .. so we are in a certain maqam in a certain root note, then the little sound effects play a nice little song .. this in itself could be satisfying and addictive."*

## What is wrong today, counted

| | |
|---|---|
| Games whose cues are a table of fixed frequencies | **all 28** |
| Games with a white-noise buffer source | **24 of 28** |
| Distinct pitches from ten taps on THE STEELYARD's ones plate | **one** |

11,064 lines of audio code across the fleet, none of it shared except the
safety bus. The problem is not the timbre and it is not the volume — it is that
**nothing that happens changes what the next sound is**, so the tenth blow is
bit-for-bit the first. That is a state problem, and it is fixed in ~500 lines of
pure arithmetic with no Web Audio in them.

## What was built

`packs/shared/game-soundscape/` — 45 tests, no Web Audio, no DOM, runs in Node.

- **38 modes** as exact cents above the tonic (16 Western, 10 Hindustani thaats,
  12 Arabic maqamat), ported from beatlounge. Cents and not semitones because
  Rast's third is a neutral interval near 350¢ — rounding it deletes the maqam
  and substitutes Ionian.
- **A soundscape is four numbers**: `{ modeId, rootHz, seed, tension }`.
- **A walker**: one scale degree of state. `+1` ascends the mode, `−1` descends
  it, and after 3–8 steps the phrase resolves onto a resting degree. That
  cadence is the part that makes it satisfying rather than a scale exercise.
- **Games cannot pick pitches.** Eight gestures: `step{direction, weight}`,
  `success`, `failure`, `levelComplete`, `refuse`, `arrive`, `moreTension`,
  `lessTension`. THE STEELYARD's whole musical vocabulary is now `tune.ts`, 45
  lines including comments.
[...]

PR #780, opened 2026-08-05
dynawalla/soundscape: the groove EVOLVES — a tethered walk, steered by right and wrong, and a level turns the key over

skyl:
> *"steelyard sounds is cool, pulse is somewhat improved but it needs to gradually evolve more — the main rhythm is static. I think slowly things should evolve and change stochastically. so there is sort of the seed but it should be able to go almost anywhere. Maybe being 'right' or 'wrong' should affect the beat in different ways. We don't have to just use random but random is a good friend. Add and remove different beats with probability .. slowly evolve."*
>
> *"When does the musical mode get changed in general? Maybe it should switch more often like when a level transitions on some games."*

Closes the evolution half of #737.

---

## The thing I got wrong first, and it changed the design

`grooveMatrix` is a pure function, so PULSE got the same probabilities back
every phrase — that is the obvious reading of "static" and it is what I built
against first. Then I measured it, and **making the chart drift moved a
strike-rate profile by 0.0005.** Inaudible. PULSE's chart already varies a great
deal per phrase; the variety was already there.

The loop was the layer *underneath* it:

| | before |
|---|---|
| Bass | `BASS_PATTERNS[tier]` — a written array of beat offsets. **One** pattern, every bar, forever |
| Shaker | every offbeat eighth, every bar, forever |
| Arp | `(k + bar) % 3` — a three-bar loop |

That is "the main rhythm", and it was a loop in the literal sense. The tier now
says how BUSY the bass is and the groove says WHERE it sits — the same division
of labour `density` and the matrix already had for the player's notes. The
shaker is deliberately left steady: something has to be the frame, or a bass
that moves reads as the tempo wobbling.

## Job 1 — the walk (`evolve.ts`, new, 79 tests in the module)

One number per instant of the bar, `-1..1`, on top of what the mode said. Every
four bars a few of them move. Three constraints keep it a groove rather than
noise, **each verified by deleting it and watching a test fail**:

1. **The metre is never drifted.** The walk moves the mode's *affinity*, never
   the metric weight, so a downbeat outranks a halfway point outranks a
   sixteenth however far it travels. The downbeat is not in the walk at all.
2. **The walk is tethered** (`SEED_HALF_LIFE_BARS = 144`) — mean-reverting, not
   Brownian. *"There is sort of the seed but it should be able to go almost
   anywhere"* is exactly that shape: a wide stationary spread around a home it
   keeps falling back through.
3. **Nothing lands mid-phrase.** Gestures queue; `matrix()` is frozen between
   `advance()` calls. A bar a child is two beats into cannot become a different
   bar.

`groove.ts` gained a `GrooveBias` seam and `leanAffinity`, which moves an
instant *inside the range the mode already allowed it* — `+1` is exactly `1`,
`-1` is exactly `MIN_AFFINITY` — so a drifted matrix passes every invariant an
undrifted one does **with no clamp anywhere**. Pass no bias and `grooveMatrix`
is byte-for-byte the pure function it always was.

**The rate is in BARS.** No tempo primitive was added; the module still has
none.

## Job 2 — right and wrong, in two directions rather than at two volumes

* **Right ADDS.** `agree()` lights up an instant the bar had been ignoring — the
  one with the most room the metre would most like to hear. **It adds no
  notes**: the budget renormalises, so the expected count is *bit-identical* to
  what the stage asked for and only WHERE changed. Rewarding correctness with a
  busier bar would be rewarding it with a harder game.
* **Wrong REMOVES.** `makeRoom()` takes the busiest *decoration* down — never a
  beat, never the downbeat — and leaves a little space for a few phrases. A band
  backing off while somebody thinks. It expires on the clock and not on merit,
  so nothing has to be earned back, and it is capped at a quarter of the budget
  so playing badly on purpose never buys an easier game.

Nothing here is red, nothing is a buzzer, nothing changes the tempo. In PULSE it
is bound to **gates** (the arithmetic) and not to notes: thinning a groove
because a child dropped a hat would be the music running a critique of their
hands.

**Two designs were built and measured before this one**, both trying to make
"right" mean *the groove commits to what it is already doing*:

| attempt | result |
|---|---|
| push the instant with the highest `p` | beat an untouched groove **38 of 60** — a coin. `leanAffinity` is range-relative, so the strongest instant is by definition the one with the least room to get stronger |
| draw the target ∝ `p³` | **30 of 60**. Six pushes landed on five instants and cancelled |

Adding has headroom by construction, and it is what the brief literally asked
for.

## Job 3 — when the mode changes

**A level transition rotates the key, and the gesture already existed.**

The old rule — *never rotate while a pack owns it* — was protecting a **reason**,
not a possession. Its reason, verbatim from the code, is that *"a key change
underneath a child — mid-question, because a timer went off — is the jarring
thing"*. A transition is not that: it is the pack saying **the child finished
something and nothing is in front of them**, which is exactly the condition a
doorway satisfies. So the rule is restated rather than broken:

> **The key never moves unbidden.** Doorways are the host's boundary;
> transitions are the pack's.

The gesture is **`session.transition`** — the SDK's existing ungated session
method, whose documented contract is already the guarantee needed: *"a
transition is a thing the child finished, never a thing that beat them."* **No
wire field, no capability, no method, no `SDK_VERSION` bump.** A pack with no
levels sends nothing and rotates at doorways alone — the policy that was already
there. **Nothing in the fleet is obliged to change.**

| | |
|---|---|
| Fresh key every launch | unchanged |
| New key every `ROTATION_MS` (8 min) at a doorway | unchanged — **kept as the backstop** |
| Never moves under a playing child on a clock alone | unchanged, still a module invariant |
| **A `transition` may turn it over after `TRANSITION_ROTATION_MS` (90 s)** | **new** |
| Never the same mode twice running | unchanged, now enforced on the transition path too |

**Why the transition floor is shorter, not longer.** A doorway happens whenever
a child browses, so it needs a long floor. A transition is a real boundary a
game *declared* — a stronger signal — so it justifies a change on less
accumulated time. 90 s is long enough that a key gets to be a key even in a game
whose levels are forty seconds long.

**The 8-minute timer stays** and becomes the backstop rather than the driver:
most of the fleet has no levels, and without it a forty-minute session at THE
STEELYARD sits in one key for forty minutes.

**Guards, both tested.** A transition from a pack *not* on the stage is ignored
(React renders the incoming pack before the outgoing one's cleanup, so a message
from a dead session is a real frame). A backwards clock rotates once and settles,
as the doorway path does.

A rotation lands as a **modulation**: `Groove.retune` queues the new key to the
next phrase boundary and keeps the drift across it, exactly as `Melody.retune`
keeps the walker's degree.

## Numbers

**Drift over time** (60 seeds, eighths grid; mean change in a slot's strike
probability against the seed matrix):

| 1 phrase | 1 min | 2 min | 9 min | 35 min | 70 min |
|---|---|---|---|---|---|
| 0.012 | 0.036 | 0.050 | 0.068 | 0.073 | 0.065 |

It goes somewhere over minutes and then stays inside its spread for an hour. The
worst-affected instant moves its strike rate **0.20–0.34** — an instant that
fired in 40% of bars now firing in 10% or 70%.

**Tether.** 20 right answers reach **+4.11** of total lean; 200 bars later, with
the child answering nothing, **15%** is left. With the decay deleted, 37%
survives — which is how that assertion is known not to be vacuous.

**Right against wrong** (200 seeds per grid, 8 answers, net instants opened
minus closed):

| grid | right | wrong | right > wrong |
|---|---|---|---|
| quarters `[1]` | +1.54 | −1.31 | 196/200 |
| eighths `[1,2]` | +2.12 | −2.63 | 200/200 |
| triplets `[1,3]` | +2.35 | −3.66 | 200/200 |
| everything `[1,2,3,4]` | +1.61 | −3.93 | 200/200 |

Right leaves the expected note count **bit-identical** in 60/60 seeds, every
grid.

**Reproducible.** Same seed replays a session exactly; **60/60** different seed
pairs diverge.

**In PULSE**, where he heard it:

| | before | after |
|---|---|---|
| Distinct bass bars in 64 | **1** | **36.1** |
| Strike-rate shift, two 200-bar windows, walk **off** | — | 0.031 |
| Strike-rate shift, two 200-bar windows, walk **on** | — | **0.053** |

The A/B is the same generator with `advance` stubbed. The window has to be 200
bars: sampling noise in a strike rate falls as `1/√N` and the drift does not, so
at 96 bars the two are the same size — the first version of that test asserted a
ratio of 1.16 and failed honestly.

## Verification

`game-soundscape` 79 · `pulse` 153 · `counterweight` 165 · `dynawalla-app` 424 —
**five consecutive runs each**, plus `npm run -s tsc` on all four.
`node packs/build.mjs pulse counterweight` builds and passes the SDK checker.

**Every new assertion was mutation-tested** — broken, watched to fail, restored.
Sixteen mutations across the module, the host and PULSE. **Three did not bite
and were fixed rather than kept**, which is the part worth reading:

| mutation | what happened |
|---|---|
| delete the tether (`decay = 1`) | **PASSED.** Counting seed-crossings and bounding the hour-long excursion are both satisfied by the `±1` clamp alone. Replaced with a signed-total test that the wander (zero-mean) cannot fake: fails at 37% vs a 25% threshold |
| always one mutation per event, ignoring grid width | **PASSED.** The grid-rate bound was `0.5×–2×`; the real ratio is 0.96 and the mutation's is 0.68. Tightened to `0.8×–1.25×` |
| drop `groove.revision` from the phrase cache key | **PASSED — and the comment claiming it was load-bearing was wrong.** Queue-then-apply is what stops a bar changing under a child; the key is defence in depth for a future caller. Comment corrected, and the test now bites on the real guarantee (fails in 1 ms if `agree()` starts landing immediately) |

Also found by mutation testing: `rig("wrong", 4)` **never actually stumbles** — a
bot that drums perfectly earns back more health than a wrong gate costs — so "a
stage LOST is not a transition" was vacuous. The rig grew a `"silent"` bot that
genuinely falls. And a failing assertion left a `Run` undisposed, holding the
test file open for Node's full 60 s timeout instead of printing one line of red;
the new tests use `try/finally`.

## Seams left for FORGE and TREBUCHET

* `Groove` is optional — a game with no bar clock does not need one. `Melody` is
  untouched and every existing gesture behaves exactly as before.
* `grooveMatrix(scape, spec)` with no bias is the same pure function it was.
* `bandBeats` in PULSE is the pattern for "a backing layer drawn from the groove
  rather than written down", and it is thirteen lines.
* The walk's universe is the **union** of every grid it is shown, so two layers
  in one game may ask about different subdivisions **in either order** without
  changing the music — tested, because `scheduleBacking` runs before the player
  notes are laid and moving one line would swap them.

`SOUNDSCAPE_DESIGN_2026-07.md` carries the full policy and the rejected designs.

### DIFF
--- a/dynawalla/docs/SOUNDSCAPE_DESIGN_2026-07.md
+++ b/dynawalla/docs/SOUNDSCAPE_DESIGN_2026-07.md
@@ -1,6 +1,6 @@
 # The Dynawalla soundscape
 
-**Status:** live. `packs/shared/game-soundscape/` exists and has 79 tests; the host
+**Status:** live. `packs/shared/game-soundscape/` exists and has 80 tests; the host
 now chooses a soundscape and publishes it, so THE STEELYARD plays through it in
 production. **Stage 2 is now partly built** — games talk back through
 `session.transition`, and correctness shapes the beat (see *The groove evolves*,
@@ -232,6 +232,30 @@ instant of the bar and an instant means the same thing in the new mode. So a
 shape a child earned in one key is still their shape in the next one, and a
 rotation lands as a modulation of the groove that was playing rather than as one
 groove stopping and a different one starting.
+
+### Two regressions caught in review, and what they teach
+
+Both were introduced by this change and both were found by *measuring* rather
+than by listening, which is the only reason they were found at all.
+
+**The arp started playing every bar line.** `grooveMatrix` forces beat 0 to a
+certainty — correct, it is not negotiable — so a backing layer that reads the
+matrix straight lands on the bar line in every bar. The hand-written arp landed
+there in one bar of three; the new one landed there in **300 of 300**, stacking a
+pluck onto the bass and onto the chart's own downbeat. A thicker transient and a
+stiffer bar line: the exact opposite of the point. `bandBeats` now takes
+`downbeat: "keep" | "leave"` — the bass keeps it because the bass *is* the
+pulse, the arp leaves it — and a layer that stands back has beat 0's share of
+the budget handed back to it so it does not silently lose a note a bar. The
+lesson: **a certainty in the matrix is a certainty for every layer that reads
+it**, and only one layer should own the bar line.
+
+**A burst of right answers landed on one instant.** `step()` snapshotted the
+field once and spent every queued gesture against it, so four right answers in
+one phrase all scored the same instant highest and piled four steps onto it —
+saturating one beat instead of opening four, against the plural in the brief and
+against this file's own prose. The field is now re-read between gestures, capped
+at `MAX_PENDING`. Measured: one right answer lifts 1.8 instants, four lift 3.0.
 
 ### Seams left for FORGE and TREBUCHET
 
--- a/dynawalla/docs/SOUNDSCAPE_DESIGN_2026-07.md
+++ b/dynawalla/docs/SOUNDSCAPE_DESIGN_2026-07.md
@@ -1,10 +1,16 @@
 # The Dynawalla soundscape
 
-**Status:** live. `packs/shared/game-soundscape/` exists and has 57 tests; the host
+**Status:** live. `packs/shared/game-soundscape/` exists and has 79 tests; the host
 now chooses a soundscape and publishes it, so THE STEELYARD plays through it in
-production. Stage 2 (games talking back) and stage 3 (the host's ambient bed) are
-still proposed. The founder's answers to the open questions are recorded at the
-end.
+production. **Stage 2 is now partly built** — games talk back through
+`session.transition`, and correctness shapes the beat (see *The groove evolves*,
+below). Stage 3 (the host's ambient bed) is still proposed. The founder's answers
+to the open questions are recorded at the end.
+
+**The groove is no longer a fixed shape.** `evolve.ts` is an eighth file: a slow,
+tethered random walk over the bar that right and wrong answers steer in opposite
+directions, and a rotation policy that lets a LEVEL TRANSITION turn the app's key
+over. See *The groove evolves* below.
 
 **Second pack wired, and the module grew a seventh file.** PULSE reads the same
 soundscape as TIME rather than as pitch. The founder's brief for it — *"some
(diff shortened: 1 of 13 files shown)


skyl:
## Two regressions fixed after adversarial review

Both introduced by this branch, both found by **measuring** rather than by listening — which is the only reason they were found. Pushed as `263cf333d`.

### The arp was playing every bar line

`grooveMatrix` forces beat 0 to a certainty — correct, it is not negotiable — so a backing layer that reads the matrix straight lands on the bar line in **every** bar. The hand-written arp landed there in one bar of three; the new one landed there in **300 of 300**, stacking a pluck onto the bass and onto the chart's own always-present downbeat. A thicker transient and a stiffer bar line: the exact opposite of what this branch is for.

`bandBeats` now takes `downbeat: "keep" | "leave"`. The bass keeps it, because the bass *is* the pulse; the arp leaves it. A layer that stands back has beat 0's share of the budget handed back in proportion, so it keeps the note count it was asked for rather than quietly losing one a bar.

| | before this branch | first draft | now |
|---|---|---|---|
| Arp on the downbeat | 1000/3000 | **3000/3000** | **0/3000** |
| Arp notes/bar | 2.667 | 2.670 | 2.658 |
| Bass on the downbeat | 3000/3000 | 3000/3000 | 3000/3000 |
| Bass notes/bar | 3.000 | 3.015 | 3.015 |

New test asserts all four, and fails if either layer forgets which kind it is.

### A burst of right answers landed on one beat

`step()` snapshotted the field once and spent every queued gesture against it, so four right answers inside one phrase all scored the same instant highest and piled `4 × COMMIT_STEP` onto it — **one beat saturated instead of four opened**, against the plural in the brief (*"add … different beats"*) and against the file's own prose. The field is re-read between gestures now.

| right answers in one phrase | 1 | 2 | 3 | 4 |
|---|---|---|---|---|
| instants lifted | 1.8 | 2.5 | 2.8 | **3.0** |

New test; it fails against the snapshot version. `MAX_PENDING` caps queued gestures the way `advance` already capped banked bars, since each one now costs a matrix build.

### Also noted and deliberately not changed

**PULSE now sends `session.transition`, which is also the day-pass stopping point**, and its first stage climb is ~36 s in. Flagging rather than deciding: the day pass is a monetization surface and a music change should not move it by side effect. Today it is inert — nothing wires the billing installer in the shipped app, so the verdict is `play-on` — and LATTICE, COUNTERWEIGHT, FORGE and TRUEDRAW already send transitions on a comparable cadence. **If PULSE should not offer a stopping point that early, that is a one-line change and I would rather it be a deliberate yes than an accident of this PR.**

**Tier-0 bass is still one fixed hit per bar** (`BASS_PATTERNS[0]` is `[0]`, so the budget is exactly 1.0 and only the downbeat can fire). That is faithful — the ladder says one bass note at tier 0 and the downbeat is where it goes — but it means the opening ~76 s of bass sounds as it did. The arp and the chart carry the variety there.

### Re-verified

`game-soundscape` 80 · `pulse` 154 · `counterweight` 165 · `dynawalla-app` 424 — five consecutive runs each, plus `tsc` on all four. `node packs/build.mjs pulse counterweight` builds and passes the SDK checker. Headline numbers unchanged: right +1.54…+2.35 against wrong −1.31…−3.93, right > wrong in 196–200/200; 36.1 distinct bass bars in 64; strike-rate shift 0.031 walk off against 0.053 on.

